PR #14, opened 2015-02-11
Three small fixes (build error, build warning, test input)

wuestholz:
This fixes three small issues: a build error, a build warning, and an absolute path in a test input.


FrancescoLogozzo:
Thanks!

PR #21, opened 2015-03-15
Add some missing system contracts; fix bad contracts in Application and ...

tom-englert:
...DependencyObject.


Microsoft.Research/Contracts/MsCorlib/System.Collections.ObjectModel.Collection.cs
```diff
@@ -0,0 +1,394 @@
...
+        // Summary:
+        //     Gets the number of elements actually contained in the System.Collections.ObjectModel.Collection<T>.
+        //
+        // Returns:
+        //     The number of elements actually contained in the System.Collections.ObjectModel.Collection<T>.
+        public int Count
+        {
+            get
+            {
+                return 0;
```
aarondandy: Should this maybe throw a NotImplementedException?

Microsoft.Research/Contracts/PresentationFramework/Sources/System.Windows.Controls.Panel.cs
```diff
@@ -43,10 +43,11 @@ abstract public partial class Panel : System.Windows.FrameworkElement, System.Wi
     #region Methods and constructors
     protected virtual new UIElementCollection CreateUIElementCollection(System.Windows.FrameworkElement logicalParent)
     {
-      return default(UIElementCollection);
+        Contract.Ensures(Contract.Result<UIElementCollection>() != null);
+        return default(UIElementCollection);
     }
 
-    protected override System.Windows.Media.Visual GetVisualChild(int index)
+      protected override System.Windows.Media.Visual GetVisualChild(int index)
```
aarondandy: Looks like strange whitespace here

Microsoft.Research/Contracts/PresentationFramework/Sources/System.Windows.FrameworkElement.cs
```diff
@@ -54,8 +54,9 @@ protected sealed override void ArrangeCore(Rect finalRect)
     {
     }
 
-    protected virtual new Size ArrangeOverride(Size finalSize)
+    protected virtual Size ArrangeOverride(Size finalSize)
     {
+      Contract.Requires(!finalSize.IsEmpty);
```
aarondandy: I can't find anything backing this one up. Can you add a comment for it?

Microsoft.Research/Contracts/PresentationFramework/Sources/System.Windows.FrameworkElement.cs
```diff
@@ -138,8 +139,9 @@ protected sealed override Size MeasureCore(Size availableSize)
       return default(Size);
     }
 
-    protected virtual new Size MeasureOverride(Size availableSize)
+    protected virtual Size MeasureOverride(Size availableSize)
     {
+      Contract.Requires(!availableSize.IsEmpty);
```
aarondandy: I can't find anything backing this one up. Can you add a comment for it?

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.Expression.cs
```diff
@@ -5566,8 +5566,14 @@ public sealed class Expression<TDelegate>
   {
     private Expression() { }
 
-    // public TDelegate Compile();
+#if !SILVERLIGHT
+        public TDelegate Compile()
+        {
+            Contract.Ensures(!ReferenceEquals(Contract.Result<TDelegate>(), null));
```
aarondandy: Probably right but I am curious, why not `Contract.Result<TDelegate>() != null` ?

Microsoft.Research/Contracts/System/System.Collections.ObjectModel.ObservableCollection.cs
```diff
@@ -0,0 +1,193 @@
...
+        // Summary:
+        //     Raises the System.Collections.ObjectModel.ObservableCollection<T>.CollectionChanged
+        //     event with the provided arguments.
+        //
+        // Parameters:
+        //   e:
+        //     Arguments of the event being raised.
+        protected virtual void OnCollectionChanged(NotifyCollectionChangedEventArgs e)
+        {
+            Contract.Requires(e != null);
```
aarondandy: This makes sense to me but I don't see anything that would indicate it is required. Could you justify with a comment maybe?

Microsoft.Research/Contracts/System/System.Collections.ObjectModel.ObservableCollection.cs
```diff
@@ -0,0 +1,193 @@
...
+        // Summary:
+        //     Raises the System.Collections.ObjectModel.ObservableCollection<T>.PropertyChanged
+        //     event with the provided arguments.
+        //
+        // Parameters:
+        //   e:
+        //     Arguments of the event being raised.
+        protected virtual void OnPropertyChanged(PropertyChangedEventArgs e)
+        {
+            Contract.Requires(e != null);
```
aarondandy: This makes sense to me but I don't see anything that would indicate it is required. Could you justify with a comment maybe?

Microsoft.Research/Contracts/MsCorlib/System.Collections.ObjectModel.Collection.cs
tom-englert: I don't think this would make any difference, CC will only look at the contracts information and ignore any code.

Microsoft.Research/Contracts/WindowsBase/System.ComponentModel.ICollectionView.cs
```diff
@@ -0,0 +1,505 @@
...
+                Contract.Ensures(Contract.Result<ObservableCollection<GroupDescription>>() != null);
+                throw new NotImplementedException();
+            }
+        }
+
+        ReadOnlyObservableCollection<object> ICollectionView.Groups
+        {
+            get
+            {
+                Contract.Ensures(Contract.Result<ReadOnlyObservableCollection<object>>() != null);
```
aarondandy: The documentation mentions that this could return null if there are no groups: https://msdn.microsoft.com/en-us/library/system.componentmodel.icollectionview.groups(v=vs.110).aspx

Microsoft.Research/Contracts/MsCorlib/System.Collections.ObjectModel.Collection.cs
aarondandy: I'm more curious here. I prefer to return as I think it is lighter but was wondering it it made any difference. I suppose it does not matter if we have a standard, not yet at least.

aarondandy:
I'm done reading up to 227167b042b5657b98e34736b15bebb6123e4b08 . Some good changes in this PR. Found some stuff I was unsure about but only found one item that I think need to be addressed: https://github.com/Microsoft/CodeContracts/pull/21/files#diff-cace338ba5980de8d18370a9d6b804e2R389

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.Expression.cs
tom-englert: Because there is no constraint that TDelegate is a class, so for the compile it could be a value type, too - and you cant do value_type != null

Microsoft.Research/Contracts/PresentationFramework/Sources/System.Windows.Controls.Panel.cs
tom-englert: Sorry, it's hard to keep the old formatting with 2 spaces indent where the default in VS is 4 spaces. Are there any coding guidelines for CC project?

Microsoft.Research/Contracts/PresentationFramework/Sources/System.Windows.FrameworkElement.cs
tom-englert: I'll add some comments in the code.

Microsoft.Research/Contracts/PresentationFramework/Sources/System.Windows.FrameworkElement.cs
tom-englert: I'll add some comments in the code.

Microsoft.Research/Contracts/System/System.Collections.ObjectModel.ObservableCollection.cs
tom-englert: Move has the same constraints as Remove and Insert, index must always be within bounds of the list.

Microsoft.Research/Contracts/System/System.Collections.ObjectModel.ObservableCollection.cs
tom-englert: CC requires this for all event handlers: 
If I connect to an event and check e for null in the event handler, CC will complain that this is redundant.

Microsoft.Research/Contracts/WindowsBase/System.ComponentModel.ICollectionView.cs
tom-englert: Then this documentation might be wrong. Generally plain properties returning ObservableCollections must never change the returned object, since you need to connect to the returned objects CollectionChanged event to wait for some object to arrive - which would be impossible if null is returned. 
Also since this is a read only property, we could not set a new collection but only add items to the existing collection - which is also impossible if null would be returned.

Microsoft.Research/Contracts/PresentationFramework/Sources/System.Windows.Controls.Panel.cs
mike-barnett: I think you should feel free to change the convention to 4. Maybe just open an issue to ask for a vote. But it would be easier for everyone if it matched the VS standard.

Mike

Sent from my Windows Phone
________________________________

Microsoft.Research/Contracts/System/System.Collections.ObjectModel.ObservableCollection.cs
aarondandy: My comment was meant for OnCollectionChanged but you answered it bellow, thanks.

Microsoft.Research/Contracts/MsCorlib/System.Collections.ObjectModel.Collection.cs
tom-englert: Most of the stuff is auto-generated, and it would generate a huge effort to change everything, with low benefit.

aarondandy:
:+1:

hubuk:
Looks good :+1:

sharwell:
:exclamation: I think this pull request needs to be closed. Changes pushed to **tom-englert/master** after this pull request was created were automatically appended to this pull request, and I don't think that was intended.

tom-englert:
It was intended for my changes - however everything is already collected together by "hubuk". As long as my changes make it to the main branch, I'm fine with this.

sharwell:
I partially reviewed this pull request so far. The first two things I note are:

1. The pull request has definitely been updated at this point to no longer include the changes mentioned in my previous comment.
2. This pull request is large; the biggest challenge is it contains all of 1) items that *need* to be merged because they fix errors in the code base, 2) items that would be nice to merge because they improve the quality of the code base, 3) items which I am unsure about at this point, and 4) items which I would prefer not to merge as part of this pull request.

Concern was expressed in #62 regarding conflicts in this pull request. I am not very concerned about this right now for the following reasons:

1. The current conflicts (the reason GitHub doesn't want to merge it) are not a problem for a manual merge of this pull request. @tom-englert would not need to make any changes to this pull request in order for a successful merge to be made.
2. The ongoing pull requests for build improvements don't overlap or conflict with the primary improvements in this pull request, so it will be relatively easy to prevent new conflicts from arising.

As long as this pull request gets merged before #63 is merged, I believe we are fine.

:question: @tom-englert I've been focused on getting the standard build process updated with the new tooling required for IL rewriting to support the Visual Studio 2015 build tools, and for the editor extensions to support Visual Studio 2013. With the understanding that I really want to get this merged, would you be OK with waiting for me to do a more extensive code review after the build process is updated to account for foundation items like #57, #60, and #66?

tom-englert:
@sharwell sounds good to me, as long as the changes will make it to the next release I'm fine with everything.

The new solution is just for convenience, it dramatically speeds up development when fixing the existing contracts. It would be helpful to anyone concentration on contract definitions only, but I won't insist in keeping this in the repo.

I tried my best to not touch the formatting, which is a very hard task with the existing code base without switching of all automatism in all tools. I think there is already an issue requesting to standardize formatting of the existing code, so anyone can work with standard VS settings, which I assume most developers already use. Until this is done I think we have to live with minor formatting and white space changes.

tom-englert:
I have split this into https://github.com/Microsoft/CodeContracts/pull/83 and https://github.com/Microsoft/CodeContracts/pull/84. Both are now based on the latest master and have reduced white space changes, so they can be automatically merged. All reviews from this request should still apply.

Issue #18, opened 2015-03-07
Support for VS2015


thegreatco:
Can we add support for VS2015? I use it for most of my projects but need to keep 2013 around for projects that use CodeContracts.

aarondandy:
I wonder if this would be a large undertaking as the compiler output from Roslyn may be different enough to cause problems. I think 2015 support is going to be very important to keep this project going but it seems like we are going to be the ones that need to do it. Hopefully somebody can get enough time to dig in and learn the code base.

Daniel-Svensson:
It seems code contracts works ok with VS 2015 but there are a few issues to fix.
This is the result om my research, hopefully if will make adding support for VS 2015 much easier.

* It does not work with code using "dynamic" since roslyn uses new metadata embedded in the pdb files.
In order to get it to work with roslyn it seems like the CCI project needs to be updated to use the latest version (if it has not already been so) .
http://dissipatedheat.com/2014/10/28/curious-case-of-code-contracts-and-roslyn/
 http://blog.xoc.net/2014/11/code-contracts-and-visual-studio-2015.html
https://roslyn.codeplex.com/workitem/54

* Target (and probably other) files installed into version specific folders need to be added at the correct corresponding locations for VS 2015
Target files which are now only copied to C:\Program Files (x86)\MSBuild\12.0\Microsoft.Common.Targets\ImportAfter\ should be copied to C:\Program Files (x86)\MSBuild\14.0\Microsoft.Common.Targets\ImportAfter\ for VS 2015
http://blog.xoc.net/2014/11/code-contracts-and-visual-studio-2015.html

* The tooling /settings tab (GUI) should be updated to work inside VS 2015

roji:
+1, thanks for doing this work!

RxDave:
+1
Looks like the CCI bug was fixed already: 
https://ccimetadata.codeplex.com/SourceControl/changeset/73073

wasabii:
Mine is actually throwing an error on 7. ;)

Daniel-Svensson:
Updated my comment above, seems like the first big issues has been fixed but immediately found a new issue,

There might be some more cases where the rewriter will miss contracts but at least it is now possible to fetch the latests source (I merged the currently pending pull requests) and get started.

erwinwolff:
Keep up the good work! I am also working with code contracts on a large project and I have to see them go.

danielcweber:
[...]

Issue #30, opened 2015-04-15
MSBuild 12.0 Microsoft.CodeContracts.targets imports wrong ContractAnalysis.targets


jbcutting:
The MSBuild v12.0 version of Microsoft.CodeContracts.targets imports $(CodeContractsInstallDir)MsBuild\v4.0\Microsoft.CodeContractAnalysis.targets. This is a relatively minor issue, but causes some problems if you're trying to add the minimum requirements for Code Contracts into your source enlistment to support builds without the extension.

The problem is on line 638 with the definition of the CodeContractAnalysisTargets property. This line is fine in the MSBuild 3.5 and 4.0 versions of the file.

The best way to resolve this would be to set the property to $(MSBuildThisFileDirectory)Microsoft.CodeContractAnalysis.targets, which would prevent the need for the version number in any of the files (4.0 and up - 3.5 doesn't support MSBuildThisFile).

Most of the content could be consolidated to avoid the need for duplicating the file with each version, but this minor change would avoid at least one problem.

hubuk:
For now I would prefer a minor change. But, could you explain in what scenario current target path is causing problems? Is this related to build server workspace? I just want to know the nature of the problem.

jbcutting:
It's not an issue if the entire MSBuild folder is there, but when adding to source control, it's preferable to reduce things to what's needed. Since the v12.0 file references the v4.0 folder, both folders have to be there. This would also be an issue if the 4.0 folder were removed at some point.

The simple change is to do the following, which avoids the need to worry about the version at all. This can be applied to the 4.0, 12.0, and 14.0 .targets files.

<CodeContractAnalysisTargets>$(MSBuildThisFileDirectory)Microsoft.CodeContractAnalysis.targets</CodeContractAnalysisTargets>

sharwell:
Fixed by #57.

PR #36, opened 2015-04-28
Updated the setup to support installing to Visual Studio 2015.

billings7:
Partially addresses #18

Full support for 2015 still needs CCI to be updated to support the custom metadata added by the Roslyn compiler when `dynamic` is used.


aarondandy:
Is anybody watching this PR with access to upload a new installer?

VSadov:
Re: needs CCI to be updated to support the custom metadata added by the Roslyn 

As I understand the most recent version of contract tools should already have updated CCI that has fixes to handle Roslyn outputs. I am not familiar with this codebase though, so I do not know where CCI lives in the tree.

cc @tmat

billings7:
The assemblies in Microsoft.Research\ImportedCCI2 are what needs to be updated. I could do it, however I can't seem to find the source for a couple of the assemblies, AsmMeta.exe an CCDoc.exe.

vossad01:
With VS2015 having released an RC build, it would be great for a build with this to be released :)

hubuk:
Already recompiled Microsoft.Cci.PdbReader.dll with most recent changes from Codeplex. Commit is pending...

billings7:
I'll have to check if an updated CCI will fix it but I haven't been able to build a project I am working on in 2015, the rewriter fails with some error about not being able to resolve a type.

hubuk:
The error message with type resolving is probably related to Roslyn issue regarding compilation of VisualBasic projects with /nostdlib or /noconfig. I will check it later and raise an issue to Roslyn if confirmed.

hubuk:
:+1:  Tested on Visual Studio Enterprise 2015 RC. Seems to be working. Great work!

Daniel-Svensson:
:thumbsup: Installation seems to go fine and code contracts is run as expected under 2015 RC. So a god step on the way to full support for VS 2015

It does seems like it does not rewrite Contracts.Requires(...) inside async methods under VS 2015 (with roslyn) but under VS 2013 it works fine, but that is another problem.

hubuk:
Finally diagnosed what was causing errors about not being able to resolve a type. It was a wrong MyType for Microsoft.VisualBasic.Contracts project. Fixed in #43.
Despite some warnings about metadata version compilation under VS2015 shall be working now (with #39 #40 and #43).

billings7:
The problem I have is caused by another issue. I have a solution, just need to get around to do a pull request.

billings7:
I have fixed one problem I was having (#47, pr #48), but that has lead me on to another problem that I currently have no clue how to solve (#49)

Microsoft.Research/ManagedContract.Setup/MSBuild/v14.0/Microsoft.CodeContracts.targets
```diff
@@ -0,0 +1,654 @@
...
+    <Exec
+       Command='"$(CodeContractsCCdocgenCommand)" "@$(IntermediateOutputPath)$(ProjectName)ccdocgen.rsp"' 
+       />
+  </Target>
+
+  <!--=====================================================================
+      Include Code Analysis target if present
+    ======================================================================-->
+  <PropertyGroup>
+    <CodeContractAnalysisTargets>$(CodeContractsInstallDir)MsBuild\v4.0\Microsoft.CodeContractAnalysis.targets</CodeContractAnalysisTargets>
```
hubuk: According to issue #30 this line should point to v14.0 of the targets folder.

sharwell:
:thought_balloon: The alternative to this is to distribute the MSBuild tooling exclusively as a NuGet package starting with the Visual Studio 2015 support, and use that as a model for transitioning the older tools as well.

SergeyTeplyakov:
Guys, what is the current status of this PR?

billings7:
@hubuk suggested that the binary files should not be committed, but be created by part of the build process, which I agree with. However I feel that would be better suited for a different PR, so leave them here for now as they would be a pain for anyone else to built ATM (they were for me anyway) and so we can get an updated setup out ASAP. As long as we are fine with that, this one should be good to go.

SergeyTeplyakov:
Merged manually.

PR #39, opened 2015-05-07
Fix for issue #33

### LINKED ISSUE #33: ccrefgen uses PdbFunction.ReadCustomMetadata_OLD instead of ReadCustomMetadata
`PdbFunction.ReadCustomMetadata_OLD` throws when there is an unknown custom metadata. 
Roslyn and dynamic local variables introduce metadata types greater than 4 which causes `ReadCustomMetadata` to throw a `PdbDebugException` with the message "Unknown custom metadata item kind: {0}"

Repro at https://github.com/krk/CodeContractsRosyln

Note: ccrewrite uses the new ReadCustomMetadata, ccrefgen does not.

hubuk:
Recompiled Microsoft.Cci.PdbReader assembly using latest version (commit 73192 - Mar 23 2015 1:29 AM - version 1.0.13.0). This version contains fix commited by Mr Mike Barnett (commit 73073) for reading custom metadata.
Compiled assembly has version 0.0.0.2 to minimize changes required. It is safe from other assemblies' perspective as public interface has not changed.
This should fix #33


sharwell:
:question: Where does the source code for this dependency live? Is it, or can it be, distributed through NuGet? Tracking arbitrary versions in the code base like this tends to be problematic over time; when we update dependencies we should make the effort to update them to more reliable references.

hubuk:
Source code for CCI is located on [CodePlex](https://ccimetadata.codeplex.com/)
NuGet packages for CCI are outdated. I'll ask someone from project members to make a new release.

Issue #17, opened 2015-02-26
Code Contract Editor Extensions for VS 2012 crash on start


jterry75:
Every time I start Visual Studio with the "Code Contracts Editor Extensions VS2012" enabled it crashes on start and asks me to send the logs. I am using:

Microsoft Visual Studio Ultimate 2013
Version 12.0.31101.00 Update 4

Microsoft .NET Framework
Version 4.6.00031

Is there a more up-to-date version for VS 2013 that does not crash? I have not been able to locate it.

Thanks,
Justin

hubuk:
Shall be fixed in PR #44.

sharwell:
Fixed by #66.

PR #40, opened 2015-05-07
Replaced calls to UpdateManifestVersion with XmlPoke

### LINKED ISSUE #34: Missing UpdateManifestVersion.exe
Its not possible to build the setup by running buildCC as it expects the executable UpdateManifestVersion.exe to exist in Microsoft.Research\common\bin.

hubuk:
Recreated missing UpdateManifestVersion console application project. The root cause was a missing exclusion from .gitignore file.
This should fix #34

### DIFF
--- a/Microsoft.Research/ManagedContract.Setup/buildMSI10.xml
+++ b/Microsoft.Research/ManagedContract.Setup/buildMSI10.xml
@@ -202,21 +202,46 @@
   <Target Name="UpdateVSIXVersions"
           DependsOnTargets="Validate">
     <Message Text="Updating Version Number in VSIX manifest"/>
-    <Exec
-      WorkingDirectory="..\ContractAdornments"
-      Command="&quot;..\common\bin\UpdateManifestVersion.exe&quot; $(CCNetLabel) VS2010\source.extension.vsixmanifest" />
-    <Exec
-      WorkingDirectory="..\ContractAdornments"
-      Command="&quot;..\common\bin\UpdateManifestVersion.exe&quot; $(CCNetLabel) VS2012\source.extension.vsixmanifest" />
-    <Exec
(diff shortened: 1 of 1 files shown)


sharwell:
:exclamation: I think this should be updated to build the new project as part of the build script, and continue to exclude the compiled binary from source control.

sharwell:
:exclamation: The build script should not be directly modifying files which live in source control. Eventually this build step will need to be replaced by either a build step that copies the file to the intermediate output directory and applies the transformation there, or moved to be a separate process which is executed to update version numbers on-demand when official releases are made.

hubuk:
To be closed/modified. Will check if XmlPoke task can be used instead of an executable file.

hubuk:
Should fix issue #34.

Main weaknesses in original approach but not addressed in this PR are:
1. Version numbers are modified in place. There is no copy-to-temporary-location-and-change approach. This issue should be addressed in #65.
2. Number of vsixmanifest files should be reduced to one in most optimistic scenario. Probably will be covered in a solution for #17 or #65.

sharwell:
Yes, my solution for the VSIX distribution only uses one vsixmanifest file.

sharwell:
This is an obviously superior strategy to the previous executable. Good job on that.

My only real question is whether we want to wait for me to finish refactoring the VSIX or not. It shouldn't be too much more work.

hubuk:
I can surely wait. What would be the best way to do that? Closing this PR for now and reopen it in slightly modified form after your changes?

sharwell:
That works for me :)

hubuk:
Great. I am freezing it for now.

mike-barnett:
Thanks so much! This is really great. You are all doing an amazing job getting this code base into shape. I can't tell you how much I appreciate it.

Issue #42, opened 2015-05-13
Contract assembly for Microsoft.VisualBasic.dll should define "Empty" _MYTYPE.


hubuk:
Using C# compiler from VS2013 produces no "My" namespace in Microsoft.VisualBasic.Contracts.dll even if _MYTYPE is set to "Windows". This behavior changed in Roslyn and in order to disable My namespace in output assembly _MYTYPE shall be set to "Empty". Setting it to "Empty" in Microsoft.VisualBasic10.vbproj file will fix compilation for VS2015 and at the same time will not change compilation results for VS2013.

hubuk:
Fixed by #43

PR #43, opened 2015-05-13
Fix for issue #42

hubuk:
MyType Empty is a type which should be used to disable My namespace in VB assemblies.
This fixes issue #42 and along with other pull requests enables compilation under VS2015.


sharwell:
:+1:

Issue #4, opened 2015-01-09
Support VS2013 for CC Editor Extensions


reitzmichnicht:
Editor Extensions throw an exception when only VS2013 is installed on a machine. Please provide a new extension that supports VS2013.

Even when registering the missing VS2012 dlls in the GAC the intellisense support to show contracts doesn't work.

aarondandy:
If somebody has a machine that reproduces this could they provide a Fusion Log for the failed bind, assuming it is a failed bind.

sharwell:
This is something I can probably take a look at, but it could be a few days.

reitzmichnicht:
Microsoft.VisualStudio.Csharp.Services.Language.dll is the dll it tries to load from VS2012

reitzmichnicht:
Corresponding forum post:
https://social.msdn.microsoft.com/Forums/en-US/a79a3a31-ef22-4946-8b00-9953c2d57617/code-contracts-editor-extensions-on-vs-2013-rtm-crash?forum=codecontracts

sharwell:
@reitzmichnicht Thanks for that update. I've worked with this DLL successfully in a single VSIX which supports Visual Studio 2010 through 2015:

https://github.com/tunnelvisionlabs/InheritanceMargin

Daniel-Svensson:
@sharwell Have you made any progress with VS 2013 support?
I would definitly love support for VS 2013 (and maybe 2015)

PR #44, opened 2015-05-15
Fix for issue #4: Support VS2013 for CC Editor Extensions

hubuk:
Added referenced assemblies from VS2010 and VS2012 to fix Contracts Editor Extensions compilation. Updated VS extension projects to not trigger one-way upgrade warning.
Possible fix for issue #4.


sharwell:
I need to take a closer look at this. I have a set of NuGet packages designed specifically to address these concerns with greater assurance, all without adding the binaries to source control.

I'm :-1: on this particular approach, not because it won't work but because there is an easier, cleaner way.

I'll add more information soon.

sharwell:
@hubuk I made an initial pass through this. I plan to resolve this issue as follows:

1. Use a single project/assembly for the **Adornments** code targeting Visual Studio 2010+. This is easy since the project does not depend on any newer/incompatible APIs.
2. Use a single project/assembly for the **ContractAdornments** code targeting Visual Studio 2010+. The only difficult part here is a dependency on the C# language service which is specific to specific versions of Visual Studio. To resolve this, I plan to use exactly the same approach that I used for my [Inheritance Margin](https://github.com/tunnelvisionlabs/InheritanceMargin) extension (which depends on exactly the same set of assemblies).
  * Move the code which requires the C# language service to a set of small assemblies that are tied to specific versions of Visual Studio
  * Load the correct assembly at runtime

hubuk:
@sharwell Your approach seams to be nice and clean. I suspected that there are no differences between these projects other than referenced assemblies. This can be handled out of the box thanks to version redirections in devenv.exe.config (except laguage services as you mentioned). I don't know if Adornments* and ContractAdornments* projects are using all the declared references but the list is quite big. It is also worth mentioning that following assemblies are missing from the repsository:
Microsoft.VisualStudio.CoreUtility.dll
Microsoft.VisualStudio.Editor.dll
Microsoft.VisualStudio.Language.StandardClassification.dll
Microsoft.VisualStudio.Text.Data.dll
Microsoft.VisualStudio.Text.Logic.dll
Microsoft.VisualStudio.Text.UI.dll
Microsoft.VisualStudio.Text.UI.Wpf.dll

It would be best to have Visual Studio extension libraries accessible via NuGet but unfortunately Microsoft has not published them for dev10.

sharwell:
@hubuk Microsoft has not. [But I have.](https://www.nuget.org/packages?q=owner%3Asharwell+tag%3Avssdk)

sharwell:
This is (presumably) superseded by #66.

hubuk:
Yup, closing it now.

Issue #47, opened 2015-05-19
"Could not resolve type reference" for some iterator methods in VS2015


billings7:
Any iterator methods which are explicitly implemented from an interface, and the interface contains multiple generic type arguments, causes a "Could not resolve type reference" error in the re-writer when built in Visual Studio 2015. 
For example:
```c#
public class TestYield : IEnumerable<Tuple<int, int>>
{
    IEnumerator<Tuple<int, int>> IEnumerable<Tuple<int, int>>.GetEnumerator()
    {
        yield break;
    }

    IEnumerator IEnumerable.GetEnumerator()
    {
        yield break;
    }
}
```
fails with `Could not resolve type reference: [TestContracts]TestContracts.TestYield.<System-Collections-Generic-IEnumerable<System-Collections-Generic-KeyValuePair<System-Int32\,System-Int32>>-GetEnumerator>d__0.`

The root cause of this is the type name parser used by the re-writer to read attributes does not support reading type names with escaped characters. The Rosyln compiler is adding an IteratorStateMachineAttribute to the method pointing to a type with the name `<System-Collections-Generic-IEnumerable<System-Collections-Generic-KeyValuePair<System-Int32,System-Int32>>-GetEnumerator>d__0`, and the comma in the generic arguments gets escaped with a backslash in the IL.

Pull request incoming.

SergeyTeplyakov:
Double checked that fix worked. Closing the issue.

P.S. Still trying to add an unit test for this case...

PR #48, opened 2015-05-19
Re-wrote the parser for serialized type names.

billings7:
Fix for issue #47 

The parser is based on the equivalent parser used in CoreCLR (see https://github.com/dotnet/coreclr/blob/master/src/vm/typeparse.cpp)


SergeyTeplyakov:
Merged.

Issue #51, opened 2015-05-19
Contract.Requires not rewritten in async methods compiled by Roslyn


danielcweber:
There already are some open issues that deal with the rewriter missing Asserts/Assumes in async methods/iterators (notably #41), however, these are concerned with rather complex conditions in these Asserts.

This issue is about `Contract.Requires` with simple conditions (such as `param != null`) in async methods compiled by Roslyn (VS 2015).

Simple working example (class library project):

    using System.Diagnostics.Contracts;
    using System.Threading.Tasks;

    namespace RewriterWithRoslynRepro
    {
        public class Class
        {
            public async Task<int> AsyncMethod(string str)
            {
                Contract.Requires(str != null);

                await Task.Delay(100);

                return 0;
            }
        }
    }

After some investigation (I am not familiar with these sources), the rewriter seems to do the following steps upon encountering an async method (i.e. a method that sets up a state machine and so on)
 - it looks up the `MoveNext` method of the state machine that is generated by the compiler
 - it simulates what `MoveNext` would do when it's called for the first time
 - it finds an IL block "B" that possibly contains a `Contract.Requires` statement
 - it looks up the last `Contract.Requires` statement that occurs after or in "B". By last, I mean highest in terms of the IL offset of the statement.

The lookup of such a block "B" succeeds for IL compiled with the native compiler but fails for Roslyn compiled IL. Upon further investigation, i found [this](https://github.com/Microsoft/CodeContracts/blob/master/Foxtrot/Foxtrot/Extractor.cs#L2507) assumption in the code.

From my experiments with Roslyn compiled code, this does not hold (any more?). The initial state is still -1. Changing [this](https://github.com/Microsoft/CodeContracts/blob/master/Foxtrot/Foxtrot/Extractor.cs#L2833) line to `var initialState = isAsync ? -1 : 0;` lets the rewriter find the block.

Can someone with insight into Roslyn confirm that the initial state of an async state machine is indeed -1 and not 0 ?

EDIT: According to the Roslyn sources ([here](https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Portable/Lowering/StateMachineRewriter/StateMachineStates.cs) and [here](https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Portable/Lowering/AsyncRewriter/AsyncRewriter.cs#L173)) the initial state is indeed -1.

matteo-mosca:
[...]

PR #52, opened 2015-05-19
Possible fix for issue #51

danielcweber:
According to the Roslyn sources ([here](https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Portable/Lowering/StateMachineRewriter/StateMachineStates.cs) and [here](https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Portable/Lowering/AsyncRewriter/AsyncRewriter.cs#L173)) the initial state of a state machine that Roslyn creates for an async method is -1. Code Contracts assumes that for Roslyn, the initial state is 0 and thus fails in rewriting async methods compiled by Roslyn. See #51 for more details.

### DIFF
--- a/Foxtrot/Foxtrot/Extractor.cs
+++ b/Foxtrot/Foxtrot/Extractor.cs
@@ -2504,7 +2504,7 @@ namespace Microsoft.Contracts.Foxtrot {
     ///       - assignment to state
     ///       - unconditional branch
     ///       
-    /// Wrinkle: in Rosly, the initial async state is also 0, not -1.
+    /// Wrinkle: in Roslyn, the initial async state is also -1.
     /// The context uses the linker version to determine if this assembly was Roslyn generated.
     /// 
     /// Another wrinkle is async methods that are not really async and C# still emits a closure etc, but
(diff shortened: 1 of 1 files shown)


matteo-mosca:
I'm actually looking forward to this. This is the only thing preventing me to migrate my current work to .net 4.6 and aspnet v5.

Daniel-Svensson:
:+1: This solves my problems with async.and roslyn.

SergeyTeplyakov:
Few questions/suggestions:
1) Please update a comment and put following link as a proof that Roslyn is using the same constants as old compiler: https://github.com/dotnet/roslyn/blob/b6484300dfafb43af0c27e542ec457a7583e1aa8/src/Compilers/CSharp/Portable/Lowering/StateMachineRewriter/StateMachineStates.cs

2) Is there any changes in tests? What the impact from end-user perspective?

3) I see another place in the code with a similar logic:

Extractor.cs:GetContractClumpFromMoveNext

```c#
var initialState = moveNext.IsAsync && !isRoslyn ? -1 : 0;
moveNext.MoveNextStartState = initialState;
originalContractPosition = null;
int statementIndex;
Contract.Assume(moveNext.Body != null);
Contract.Assume(moveNext.Body.Statements != null);
int blockIndex = ContractStartInMoveNext(this.contractNodes, moveNext, out statementIndex, iteratorMethod, isRoslyn);
```

I was unable to find usages for `MoveNextStartState` but for the sake of consistency I would suggest to change this as well.

Thanks a lot for investing time in this project!

SergeyTeplyakov:
Closed as dup.

PR #54, opened 2015-05-25
Fixed usage of deafult(T) in contract abbreviators.

### LINKED ISSUE #29: Malfromed Contract. Found Requires after assignment
After abbreviators [started to work](https://github.com/Microsoft/CodeContracts/issues/23), we get the error 
>Malfromed Contract. Found Requires after assignment.

for the following method:

```C#
[ContractAbbreviator]
[SuppressMessage("Microsoft.Usage", "CA1801:ReviewUnusedParameters", MessageId = "value")]
[SuppressMessage("Microsoft.Usage", "CA1801:ReviewUnusedParameters", MessageId = "userMessage")]
public static void NotDefault<T>(T value, string userMessage)
{
	Contract.Requires<ArgumentException>(!EqualityComparer<T>.Default.Equals(value, default(T)), userMessage);
}
```

If method's body is inlined as a usual CodeContract, or abbreviator is used within the same assembly no errors appear.

**Update**
This error occurs only for enum types.

hubuk:
Should fix #29 (Malfromed Contract. Found Requires after assignment)

### DIFF
--- a/System.Compiler/Nodes.cs
+++ b/System.Compiler/Nodes.cs
@@ -7798,6 +7798,23 @@ namespace System.Compiler{
         base.DeclaringType = value;
       }
     }
+    /// <summary>
+    /// Gets a value that indicates whether the current type node is a type argument of a generic member or type.
+    /// </summary>
+    private bool isTypeArgument;
+    public bool IsTypeArgument
+    { 
+        get
+        {
+            return this.isTypeArgument;
+        }
+    }
+
+    public void MarkAsTypeArgument()
+    {
+        this.isTypeArgument = true;
+    }
+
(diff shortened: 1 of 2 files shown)


System.Compiler/Nodes.cs
```diff
@@ -7797,7 +7797,24 @@ public override TypeNode DeclaringType
       {
         base.DeclaringType = value;
       }
+    }
+    /// <summary>
```
sharwell: :bulb: this documentation comment is associated with the private field `isTypeArgument`, and the public property `IsTypeArgument` does not have any comment. This can be resolved by moving the backing field before the comment.

PR #57, opened 2015-05-27
Fixed path to Microsoft.CodeContractAnalysis.targets for MSBuild v12.0.

hubuk:
Change requested in #30 "MSBuild 12.0 Microsoft.CodeContracts.targets imports wrong ContractAnalysis.targets"

### DIFF
--- a/Microsoft.Research/ManagedContract.Setup/MSBuild/v12.0/Microsoft.CodeContracts.targets
+++ b/Microsoft.Research/ManagedContract.Setup/MSBuild/v12.0/Microsoft.CodeContracts.targets
@@ -635,7 +635,7 @@
       Include Code Analysis target if present
     ======================================================================-->
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


sharwell:
:+1: This addresses the *bug* identified in #30.

:bulb: I highly recommend creating a new issue to address the sub-optimal paths used for this property. This enhancement was mentioned in #30 and applies to each of the targets files for the various MSBuild versions.

hubuk:
I will do a rework and propose a complete solution.

sharwell:
@hubuk By the way, in case you are not familiar, you can see an overview of the emoji I use in comments and their meaning [on this page](https://github.com/openstacknetsdk/openstack.net/blob/master/CONTRIBUTING.md#emoji).

hubuk:
The change should be simple and will reduce duplicated content to maintain. I think that creating additional issue is an unnecessary overhead.

PR #61, opened 2015-06-14
Added .gitattributes

sharwell:
Many of the files in the repository were not stored according to Git conventions. This change adds an initial **.gitattributes** file and re-saves all text files using the expected line endings (`\n` in source control, `\r\n` on disk when working on Windows). The presence of **.gitattributes** will prevent inconsistencies from being introduced in future commits, regardless of the settings being used by individual developers.

The commit which re-saves files is intentionally marked as authored by @FrancescoLogozzo. This matches the original commit to this repository and ensures `git blame` does not mis-attribute lines of code to me.


yaakov-h:
FYI, `git blame` won't mis-attribute if you use `-w` to ignore whitespace.

sharwell:
@yaakov-h CodeLens in Visual Studio does not just `-w`.

sharwell:
@yaakov-h Three other places that don't use `-w`:

1. GitHub's [Blame feature](https://github.com/Microsoft/VisualStudio-TestHost/blame/71a627367865d9263a7d92afa99707e1463f6c04/BuildVSTestHost.ps1)
2. GitHub's [graphs page](https://github.com/Microsoft/VisualStudio-TestHost/graphs/contributors)
3. Most users using Git from the command line

Issue #64, opened 2015-06-16
Create a new set of lables for project.


hubuk:
What do you think about creating a new sot of labels for CC? Current one does not look consistent.
Newly created set may be similar to the one defined for [NuGet](https://github.com/NuGet/Home/labels).

sharwell:
I've been using [this set of labels](https://github.com/tunnelvisionlabs/dotnet-trees/blob/master/CONTRIBUTING.md#labels) for the StyleCopAnalyzers project. This project will certainly require a few unique labels as well:

* Missing reference contracts
* Erroneous reference contracts

sharwell:
@SergeyTeplyakov I think this can be closed. The labels are likely to evolve over time, but this is a start.

PR #66, opened 2015-06-17
VSIX rework

sharwell:
### Summary of changes

* Refactor the Visual Studio extension so a single extension works in Visual Studio 2010+
* Fully support Visual Studio 2010-2015 language services
* ~~Visual Studio 2015 support is a placeholder - the assembly exists but the service factory throws `NotSupportedException`~~ (this was completed)

Fixes #4

### Work left (should now be complete)

* ~~This is a preview branch which does not include the changes in #61. #61 needs to be merged first, and then I will update this pull request (I have a branch in my fork that's already prepared to do this efficiently).~~
* ~~Some files are missing copyright headers~~
  * :question: What is the policy on this? Ideally we could just use the [two-line comment from corefx](https://github.com/dotnet/corefx/blob/master/src/System.Collections.Concurrent/src/System/Collections/Concurrent/IProducerConsumerCollection.cs#L1-L2)
* ~~I forgot to add `[assembly: ProvideBindingPath]` (just remembered while writing this)~~
* ~~Update the build script. **buildMSI10.xml** and/or **Setup10.proj** will need major work~~

Issue #59, opened 2015-06-12
Future plans for Code Contracts


mike-barnett:
This issue is meant to provide a single place where we can keep people notified about the status of the Code Contracts project. We (Microsoft) now have some resources and will do the work (in conjunction with the community of course) to update the sources so the tools will work in VS2015. After that, we are looking to transition the repo **outside** of the Microsoft organization in GitHub and turn ownership over to any community members that are interested.

As far as we are aware, the current work items are:

1. Updating the rewriter to understand the new Roslyn code-generation patterns for iterators/async.
2. Updating the Visual Studio integration so that the Code Contracts pane appears in the project properties window.
3. Fixing some other missing files, etc.

Separately, there is a Roslyn discussion (https://github.com/dotnet/roslyn/issues/119) about adding contracts to a future version of C#.

hubuk:
There were some work done by the community:
Point 1. is partially addressed by #52.
Point 2. is resolved by #36.
Missing repository content for point 3 is listed in #6.

vmelamed:
This is great news! The only obstacle for me to start using VS2015(RC) are the code contracts. Currently almost all my projects use it. Can't wait for the release.

yaakov-h:
> ... we are looking to transition the repo **outside** of the Microsoft organization in GitHub and turn ownership over to any community members that are interested.

What's the reasoning for this?

krk:
Possibly related to the "next step" for code contracts being stated as "re-architecturing to integrate Z3" in [this document (CodeContracts.pptx)](http://research.microsoft.com/pubs/141092/CodeContracts.pptx).

PR #68, opened 2015-06-19
Add Contributing.md with a set of label descriptions.

SergeyTeplyakov:
Fix for #64.

@hubuk, @sharwell: I would appreciate for the review.


sharwell:
I too would appreciate the review :smile:

sharwell:
Looking at the issue list, I believe the project could benefit from two additional labels:

| Label | Meaning on Issue | Meaning on Pull Request |
| --- | --- | --- |
| erroneous contract | The issue concerns an incorrect contract in the reference source code | The issue concerns an incorrect contract in the reference source code |
| missing contract | The issue concerns a missing contract in the reference source code | The issue concerns a missing contract in the reference source code |

The **erroneous contract** and **missing contract** labels are the equivalent of **bug** and **enhancement** respectively, and are used when talking about contracts for the reference source code.

aarondandy:
Part of me just wants one label for a contract change, mostly because a method may have a contract that allows all input and guarantees nothing for the input. Externally for users a method that previously had no preconditions which after an update has preconditions added will appear as a contract change.

On the other hand I see benefit to the distinct labels as we don't have a good way to distinguish between a method that does not restrict input and another that was neglected. I also see the bug vs enhancement point of view but you could also apply two labels, one being contract-change and another being enhancement.

sharwell:
I would feel comfortable with a **contract** issue (or similar), where in addition to that label we can apply **bug** if the code already has a contracts but the contract is incorrect, or **enhancement** if the update is simply an improved contract representing existing behavior.

aarondandy:
Requiring those tags will probably be important for generating release notes, which is another topic that may be important.

sharwell:
@aarondandy The primary benefit I see to the **contract** tag is many contributors will feel comfortable updating reference contracts but might not feel comfortable editing the implementation. I want those users to have quick/easy access to our list of outstanding issues that they're more likely to be able to help with. :+1:

aarondandy:
Good point, I think most of those "jump-in" sites that pop up can be configured to take different labels too

SergeyTeplyakov:
It seems that some general label would be very useful. But can we come up with more descriptive name?

I mean 'contract' as a label sounds a bit weird for me. Maybe 'contract issue' or something similar?

P.S. @sharwell I'll take a look at your PR, but I don't have experience with VSIX and other black magic that you've done in #66.

CONTRIBUTING.md
```diff
@@ -0,0 +1,75 @@
+# Contributor Guide
+
+## Open Discussion
+
+The most important part of an open source project is *open discussion*. If a developer, user, or other contributor has
+a question, comment, suggestion, or concern about any part of this project, **please speak up**: create an issue or add
+a comment to an existing issue or pull request.
```
Maxwe11: Also consider create chat room in gitter.im

sharwell: Eventually that will need to happen, but I'm not sure how #59 will impact it.

aarondandy: > create chat room 

A few people here and here have been in https://jabbr.net/#/rooms/CodeContracts but most of the action is happening here in github issues.

hubuk:
I have proposed [labels for NuGet](https://github.com/NuGet/Home/labels) as a starting point because they appear to be self explanatory. Assigned colors make them easy to read and understand what set of labels are possible on single issue. Having some explanation in project wiki or other document is also a good thing. So to utilize both ideas I come with the following proposition (source markdown available on my [Wiki](https://github.com/hubuk/CodeContracts/wiki/Labels)):

## Labels

This project uses many labels for categorizing issues and pull requests. Labels are grouped according to their category:

| Label | Meaning |
| --- | --- |
| 0 - Backlog | Issue not in the scope of the current milestone. |
| 0 - Blocked | Issue blocked by some external or internal dependencies. |
| 1 - Ready | Issue is ready to be fixed. |
| 2 - Working | Someone is currently working on the issue. |
| 3 - Done | Issue already addressed and a solution has been proposed. | 
| Area: ccrewrite | Issue related to Code Contracts rewriter. |
| Area: cccheck | Issue related to Code Contracts static analyzer. |
| Area: ccdocgen | Issue related to contracts documentation generator. |
| Area: ccrefgen | Issue related to contract reference assemblies generator. |
| Area: contract | Issue related to contracts for .Net Framework assemblies. |
| Area: installer | Issue related to Code Contracts installer. |
| Area: editor extension | Issue related to Code Contracts Editor Extension. |
| Area: property page | Issue related to Visual Studio property pages for Code Contracts. |
| Area: documentation | Issue related to product documentation. |
| ClosedAs: ByDesign | The current behavior is by design and will not be changed. |
| ClosedAs: Duplicate | Another issue contains a description of the same problem. |
| ClosedAs: Invalid | The issue concerns a documented and correct behavior. |
| ClosedAs: Question | The issue is a question for which an answer has been provided. |
| ClosedAs: Fixed | The issue has been resolved. |
| Type: Bug | The issue concerns a bug in the code. |
| Type: Enhancement | The issue concerns an improvement to an existing functionality. |
| Type: Feature | The issue concerns a completely new functionality to be implemented. |

SergeyTeplyakov:
@hubuk I don't think that we need so many different labels right now. I would like to stay with minimum number of labels and add them along the way when it would be obvious that wee need all of them.

For instance, all this different labels per area would be useful only when we really would have tens of issues per each area. It could be true but could be not. The same is true for ClosedAs. Right now we have less issues than proposed labels.

As I said, I would prefer evolve those labels based on real needs. Is it ok for you?

Issue #62, opened 2015-06-16
Pull requests policies


SergeyTeplyakov:
Hello everyone.

@mike-barnett added me as an owner to this repo and I'm going to start working on pull requests and other stuff. 

My team in Microsoft is using Code Contracts library super extensively and I'm going to work on this stuff in near future to get them working with VS14, fix some other bugs, review and merge pull requests etc.

I would like to be as useful as possible and I need some help from you guys. Can you help me to prioritize and review existing pull requests? I would like to see some comments here with a list of most critical pull requests that's are ready to merge and at least one "Reviewed, looks good for me" for each pull requests. (I'll review them myself and for all critical ones I would get Mike's feedback as well).

I'm a big fan of Code Contracts and I personally don't want them to die. So I would do my best to keep the source up-to-date and merge pull requests as soon as I can.

sharwell:
You should probably make a decision on #61 (for or against) before anything else.

sharwell:
I'm actively working on updating the Visual Studio extension code to work as a single VSIX seamlessly across Visual Studio 2010 through 2015. It's very similar to one of my other open source extensions so I'm comfortable with the work. I'll have the code finished tomorrow or Wednesday; I'll send a pull request at that point even though the *installer* will be broken, and we can discuss the best approach to use for that.

SergeyTeplyakov:
@sharwell for #61: 21K of changes... It seems that this is pull-request fixes #63. 

I see that this pull-request already fixes majority of style issues but not all of them. Is it possible to run R# formatter or similar tool to fix spacing issue (use 4 spaces)? Because I would like to have one major fix but not multiple?

SergeyTeplyakov:
> I'm actively working on updating the Visual Studio extension code to work as a single VSIX seamlessly across Visual Studio 2010 through 2015.

Sound awesome!

sharwell:
@SergeyTeplyakov Aside from the addition of a **.gitattributes** file, pull request #61 *only* changes `\r\n` line endings in the repository to `\n`. Git always stores text files internally with `\n` line endings, and then uses settings of the individual developer's system to determine whether they appear as `\n` or `\r\n` when they are checked out.

SergeyTeplyakov:
@sharwell My concern that this affect so many files and maybe it would be easier to have only one commit that dramatically affects the code base. What do you think?

P.S. If you'll split the change to separate .gitattributes from \r\n I would merge the first one today.

[...]

PR #71, opened 2015-06-24
Revert "Add Contributing.md with a set of label descriptions."

SergeyTeplyakov:
Have no idea why Roslyn-based changes where applied with this pull request as well. Reverting it back.

Reverts Microsoft/CodeContracts#68


sharwell:
@SergeyTeplyakov 

1. This occurred for the same reason #21 had issues. Pull request #68 was sent from your **master** branch, so when you later updated your **master** branch the new commits were automatically appended to the pull request. I included additional information about this scenario in #62.
2. I think it would make sense to force-push to the **master** branch here to revert #68 by completely removing the commit from history. This is something we want to avoid in the future, but little (if any) work has been created on top of the current **master** branch and the repository currently contains new binary files which will never be used.

SergeyTeplyakov:
Yep, I know why this happens. I'm using different local branches for different features right now.

I thought about force-push, but decided that this could do more harm than benefits...

Issue #72, opened 2015-06-24
CCRewrite fails with NRE on async method with Ensures.ForAll with captured params because forwarder is null


SergeyTeplyakov:
There is another issue with CCRewrite when `Contract.Ensures` contains `Ensures.ForAll` with captured lambda. Related issue #69.

Following code lead to another NRE:

```c#
  class NonGen {
        public async Task<List<int>> FooAsync(int limit)
        {
            Contract.Ensures(Contract.ForAll(Contract.Result<List<int>>(), i => i < limit));

            await Task.Delay(42);

            return new List<int>{1, 2, 3, 4, 5, 6, 7, 8, 9, 10};
        }
  }
```

Trying to use this method lead to NRE at Rewriter.cs:3994:

```c#
public override Expression VisitLocal(Local local)
{
    if (HelperMethods.IsClosureType(this.declaringType, local.Type))
    {
        MemberBinding mb;
        if (!closureLocals.TryGetValue(local, out mb))
        {
            // Next line will fail, because this.forwareder is null!
            var closureField = new Field(this.closureClass, null, FieldFlags.Public, local.Name, this.forwarder.VisitTypeReference(local.Type), null);
            this.closureClass.Members.Add(closureField);
            mb = new MemberBinding(this.checkMethod.ThisParameter, closureField);
            closureLocals.Add(local, mb);

            // initialize the closure field
            var instantiatedField = GetMemberInstanceReference(closureField, this.closureClassInstance);
            this.ClosureInitializer.Statements.Add(new AssignmentStatement(new MemberBinding(this.ClosureLocal, instantiatedField), local));

        }
        return mb;
    }
    return local;
}
```

This issue is similar to #69 but the reason is different. This time the root cause is following code in Rewriter.cs (in `EmitAsyncClosure):

```c#
if (from.IsGeneric)
            {
                this.closureClass.TemplateParameters = new TypeNodeList();
                var parentCount = this.declaringType.ConsolidatedTemplateParameters == null ? 0 : this.declaringType.ConsolidatedTemplateParameters.Count;
                for (int i = 0; i < from.TemplateParameters.Count; i++)
                {
                    var tp = HelperMethods.NewEqualTypeParameter(dup, (ITypeParameter)from.TemplateParameters[i], this.closureClass, parentCount + i);

                    this.closureClass.TemplateParameters.Add(tp);
                }
                this.closureClass.IsGeneric = true;
                this.closureClass.EnsureMangledName();
[...]

Issue #70, opened 2015-06-24
Can't run Foxtrot unit tests


SergeyTeplyakov:
Maybe I'm doing something wrong, but I can't run any tests from Foxtroct/UnitTests project (VS2013, tried both - VS Test Runner and R#).

All of them are failing with an error: "Can't find the TestResults directory!!!" at RewrittenContractTests.cs:89.

`AssemblyInitialize` method should not use `Directory.CurrentDirectory` but should use `context` property of type `TestContext` to grab all the required information about TestResults folders etc.

Issue #69, opened 2015-06-23
CCRewrite fails with NRE on async method with Ensures.ForAll with captured params


SergeyTeplyakov:
Code contracts repo has a bunch of tests for async methods. Unfortunately, some of them are broken.

For instance, tests for `async6.cs` to `async9.cs` will fail because ccrewrite.exe will return -1.

Here is a small repo for the issue:

```c#
class TestTaskContracts
{
    public async Task<List<int>> FooAsync(int limit)
    {
        Contract.Ensures(Contract.ForAll(Contract.Result<List<int>>(), i => i < limit));

        await Task.Delay(42);

        return Enumerable.Range(1, 10).ToList();
    }
}
```

Running ccrewrite on this class will fail with `NullReferenceException` at System.Compiler.Writer.cs line 1814.

sharwell:
@SergeyTeplyakov Label fixed instead of up for grabs?

SergeyTeplyakov:
Done.

PR #74, opened 2015-06-24
Fix for Roslyn async methods

### LINKED ISSUE #51: Contract.Requires not rewritten in async methods compiled by Roslyn

SergeyTeplyakov:
This fix should resolve #51.

I've added latest Roslyn-based csc compiler to the tools folder and modified the tests to use it.

Unfortunately my change revealed some issues (#69, #70, #72), that should be fixed (IMO) before releasing new CC version.


mike-barnett:
Looks great! Please merge this in.

Issue #80, opened 2015-06-29
Static checker fails with "Cannot connect to the cache" on VS2015


SergeyTeplyakov:
I was able to build and run ccrewriter on VS2015, but got following failure when I tried to run cccheck:

```
CodeContracts: ConsoleApplication1: Time spent trying to connect to the cache server: 00:00:00.0156274
CodeContracts: ConsoleApplication1: Diagnostic: Failed to connect to any cache.
(1,1): warning : CodeContracts: System.Diagnostics.Contracts.ContractDeclarativeAssemblyAttribute.#ctor()[0x6]: Cannot connect to the cache. The CodeContracts static check will not run
CodeContracts: ConsoleApplication1: 
CodeContracts: ConsoleApplication1: Background contract analysis done.
```

Disableing 'Cache results' options for Static Checking fixes the issue, but by default this option turned on.

Maybe the root cause of the issue, that following dbname is used for local cache: `(LocalDb)\v11.0`.

tom-englert:
I also had frequent problems with the LocalDb cache, if the file is gone but still registered with SqlServer it's not easy to recover. 
Maybe you can have a look at https://github.com/tom-englert/CodeContracts/commit/76ed1a3350173f6fafc78f78e81ef68cf6305f21, this solved all my problems.

SergeyTeplyakov:
I solved this problem by changing connection string from (LocalDb)\11.0 to (LocalDb)\MSSQLLocalDB.

tom-englert:
Most people would be lost if the cache does not work out of the box. We should definitely improve cache handling.

SergeyTeplyakov:
I do have a fix that uses new connection string but falls back to the previous one if it fails. 

And we definitely can (and should) improve this experience! :+1:

tom-englert:
Just being curious: Where does the "MSSQLLocalDB" come from? Is the a new standard with the newer SQLServers?

SergeyTeplyakov:
Yep, this is different names for different sql server compact version.

For sql server 2012 compact is (LocalDb)\11.0 and for Sql server 2014 compact is (LocalDB)\MSSQLLocalDB.

And on VS2015 there is no 2012 version that's why old code fails.

tom-englert:
So how does this affect performance? Trying through all possible DB-Names might be a time consuming task. 
When I enter a wrong name in the SQL Management Strudio it takes about 10sec until it returns with an error. 
[...]

PR #81, opened 2015-07-02
Added an ability to specify localdb name.

SergeyTeplyakov:
Fix for #80 

This fix allows to run CCCheck on the machine with only VS2015 installed.

Please note, that new version (i.e. (LocalDb)\MSSqlLocalDb) is also available on machine with VS2013 installed. So minor performance degradation would be only for computers with ONLY VS2012 and VS2010 only.

We can enhance this strategy in the future, but this fix unblocks us to release VS2015 support (without such kind of fix CCCheck will fail unless the user manually change the options and disables caching).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -208,3 +208,4 @@ UpgradeLog*.htm
 # Microsoft Fakes
 FakesAssemblies/
 
+Microsoft.Research/CCDoc/Sandcastle/Sandcastle.zip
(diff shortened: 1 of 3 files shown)


tom-englert:
For our purpose it would be totally sufficient to look at `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Microsoft SQL Server Local DB\Installed Versions` to decide which name we can use. So we would not need that library at all.

sharwell:
Superseded by #97.

PR #83, opened 2015-07-03
Improve cache handling

tom-englert:
Deleting the data base file does no longer break CC

### DIFF
--- a/Microsoft.Research/Clousot.Cache/SQLCacheModel.cs
+++ b/Microsoft.Research/Clousot.Cache/SQLCacheModel.cs
@@ -562,7 +562,7 @@ namespace Microsoft.Research.CodeAnalysis.Caching
       get { return this.cachename; }
     }
 
-    protected static void DetachDeletedDb(string connection)
+    protected void DetachDeletedDb(string connection)
     {
       // If anyone has manually deleted the cache file, it might be still registered with LocalDB. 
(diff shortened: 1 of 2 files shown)


Microsoft.Research/Clousot.Cache/SQLCacheModel.cs
```diff
@@ -558,6 +561,34 @@ public string CacheName
...
+
+        var catalog = (string)connectionString["Initial Catalog"];
+        var dataSource = (string)connectionString["Data Source"];
+
+        using (var master = new DataContext(string.Format(CultureInfo.InvariantCulture, @"Data Source={0};Initial Catalog=master;Integrated Security=True", dataSource)))
+        {
+          master.ExecuteCommand(@"exec sp_detach_db {0}", catalog);
+        }
+      }
+      catch
```
sharwell: :bulb: I prefer to see this list the actual exception types which are known to be thrown for the cases you describe.

tom-englert: I fully agree, but here the exceptions may come from the DB providers and are neither documented nor consistent between DB providers - you can only get them by trial and error using all known providers, but then the code may break with any unknown provider.

sharwell:
I haven't tested this, but from the comments I felt I clearly understood both the problem and the way it's being solved. I'm not sure whether or not my requested change is feasible. Hopefully someone with more SQL experience can verify that the code is the best (or good enough) solution to the described problem but otherwise :+1:.

Issue #63, opened 2015-06-16
Format source code


SergeyTeplyakov:
We need to do this. Current code is not formatted very well...

sharwell:
I highly recommend we do this by adding the StyleCopAnalyzers analyzer to the project. It will both check style and offer to correct violations.

I recently added targeted support for certain alternative naming conventions. If you are interested in looking into this approach as a possibility, my suggestion would be to review the StyleCop rules and provide feedback on any rules which you foresee as a problem. For those rules, if StyleCopAnalyzers has an alternative available we can use that instead.

SergeyTeplyakov:
I'm ok with that, but I think the very first thing is to run https://github.com/dotnet/codeformatter to get this code readable. Tools for forcing this rules are super useful, but review process could mitigate this. Right now it is very inconvenient to make any fixes here.

sharwell:
I haven't used that tool; I'm not opposed to it as the first step.

I would prefer the rewrite not touch the Visual Studio extension code, as that would invalidate much of the work I've been doing. Or delay the rewrite until after I get that pull request submitted.

SergeyTeplyakov:
> I would prefer the rewrite not touch the Visual Studio extension code, as that would invalidate much of the work I've been doing.

Even more, maybe it would be better to fix VS14 first and only after that clean-up this stuff.

sharwell:
It's a no-brainer to take care of formatting for some items at the time they are added, such as the files that will be added to resolve #6. For the others, it's probably best to put a feeler out regarding any ongoing work, announce the intent to format a particular project in X days, and then if no one objects for outstanding work do the reformat. X=3 is likely to work well for people who are in different time zones and may have different weekend days.

sharwell:
The reformat should be committed with the author manually set to either @FrancescoLogozzo, @dotnet-bot, or similar in order to make sure that the entire code base isn't "hijacked" by one of the new participants. :smile:

PR #84, opened 2015-07-03
Fix various system contracts

tom-englert:
Fix a lot of buggy system contracts


Microsoft.Research/Contracts/WindowsBase/System.Windows.Threading.DispatcherObject.cs
```diff
@@ -22,7 +22,26 @@ public abstract class DispatcherObject
   {
     protected DispatcherObject() { }
 
-    public Dispatcher Dispatcher { get { Contract.Ensures(Contract.Result<Dispatcher>() != null); return null; } }
+    // DetachFromDispatcher is e.g. called by Freezable.Freeze()
+    //
+    // This method allows certain derived classes to break the dispatcher affinity
+    // of our objects.
+    // [FriendAccessAllowed] // Built into Base, also used by Framework.
+    internal void DetachFromDispatcher()
```
sharwell: :question: How does this help the contracts?

Microsoft.Research/Contracts/WindowsBase/System.Windows.Threading.DispatcherObject.cs
```diff
@@ -22,7 +22,26 @@ public abstract class DispatcherObject
...
+    {
+      // _dispatcher = null;
+    }
+
+    public Dispatcher Dispatcher
+    {
+      get
+      {
+        // Note: a DispatcherObject that is not associated with a
```
sharwell: :bulb: Looking at the source code for this class, I can confirm there are cases where this property returns `null`. However, it seems the particular case that users will recognize is `Freezable` objects which have been frozen. I recommend updating the comment to clarify this case:

> Freezable instances which have been frozen are detached from the Dispatcher, so this property will return null.

Microsoft.Research/Contracts/WindowsBase/System.Windows.Rect.cs
```diff
@@ -16,6 +16,19 @@
 using System.ComponentModel;
 using System.Diagnostics.Contracts;
 
+
+// 2015-03-36: tom-englert
+// Temporarily disabled checks for Contract.Requires(!this.IsEmpty);
```
sharwell: :exclamation: I'm not sold on this particular change. This pull request is primarily intended to add missing contracts and remove incorrect contracts, but this change is a "hack" which removes a *correct* contract. I prefer to address this (all instances where this contract was changed) by the following in order of preference:

1. Address the limitation which results in the described warning (not likely to happen).
2. Find a place where we can put `Contract.Assume` to help out the checker.
3. "Just deal with the warning", or similar
4. Move the changes to their own pull request, which can be reverted should the situation change in the future.

Microsoft.Research/Contracts/WindowsBase/System.ComponentModel.ICollectionView.cs
```diff
@@ -0,0 +1,504 @@
...
+using System.Globalization;
+
+namespace System.ComponentModel
+{
+#if !SILVERLIGHT && !NETFRAMEWORK_3_5 // => types defined in System.Windows!
+
+    // Summary:
+    //     Provides information for the System.ComponentModel.ICollectionView.CurrentChanging
+    //     event.
+    public class CurrentChangingEventArgs : EventArgs
```
sharwell: :bulb: I would prefer to see this in its own file. This isn't a big deal for now; it will likely be addressed automatically as part of #63 or a subsequent related PR.

Microsoft.Research/Contracts/WindowsBase/System.Windows.Threading.DispatcherObject.cs
tom-englert: It does not help the contracts, but to understand what's going on here any why the contract must be this way.

Microsoft.Research/Contracts/System/System.Net.WebRequest.cs
```diff
@@ -190,7 +200,18 @@ public static IWebProxy DefaultWebProxy
     //   System.NotImplementedException:
     //     If the property is not overridden in a descendant class, any attempt is made
     //     to get or set the property.
-    extern public virtual string Method { get; set; }
+    public virtual string Method
+    {
+      get
+      {
+        Contract.Ensures(!String.IsNullOrEmpty(Contract.Result<string>()));
```
sharwell: :bulb: I'll be pushing for uniformly referring to built-in types using their built-in names, which is super easy to explain to users and also super easy to automate. This comment is informational; *if* I'm successful it would be incorporated as part of #63 or a subsequent PR (e.g. to enable a subset of StyleCopAnalyzers).

Microsoft.Research/Contracts/System/System.Net.Sockets.Socket.cs
```diff
@@ -313,9 +313,6 @@ public object GetSocketOption(SocketOptionLevel optionLevel, SocketOptionName op
     public void SetSocketOption(SocketOptionLevel optionLevel, SocketOptionName optionName, object optionValue)
     {
       Contract.Requires(optionValue != null);
-      Contract.Requires((int)optionLevel == 41);
```
sharwell: :question: Were these incorrect, or just confusing?

Microsoft.Research/Contracts/System/System.Diagnostics.FileVersionInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+// The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+using System.Diagnostics.Contracts;
+using System;
+
+namespace System.Diagnostics
+{
+#if !SILVERLIGHT
```
sharwell: :question: What happens if we simply include this class when `SILVERLIGHT` is defined? If users aren't referencing this assembly directly, wouldn't everything still just work?

:bulb: When the whole type is excluded when `SILVERLIGHT` is defined, you can move the preprocessor directives to above the `using` declarations and at the end of the file, which will prevent someone from accidentally applying a Remove and Sort Usings operation that breaks the file in some context.

Microsoft.Research/Contracts/WindowsBase/System.Windows.Threading.DispatcherObject.cs
sharwell: :bulb: Thanks for clarifying. When I was reading through the code in the current form, I didn't feel like I understood exactly why `Dispatcher` could return null. If the comment for `Dispatcher` was updated in the way I mentioned in my other comment (on that property), I feel like this internal method and the preceding comment(s) could be removed from the code without reducing the clarity of this set of contracts, as freezing a `Freezable` is a well-understood concept for API consumers.

Microsoft.Research/Contracts/WindowsBase/System.Windows.Rect.cs
tom-englert: I doubt the checker will be ever capable verifying this particular case, and with this check in place code is unmaintainable, due to false positives in every line and even `Contract.Assume` does not help in many situations.

I will move this to a separate pull request and create an issue for that.

Microsoft.Research/Contracts/MsCorlib/System.Collections.Generic.IDictionary.cs
```diff
@@ -181,29 +183,31 @@ abstract class IDictionaryContract<TKey, TValue> : IDictionary<TKey,TValue>
...
     void IDictionary<TKey, TValue>.Add(TKey key, TValue value)
     {
+      Contract.Requires(!ReferenceEquals(key, null));
       // Contract.Requires(!ContainsKey(key));
       //modifies this.*;
       //Contract.Ensures(ContainsKey(key));
     }
 
-    bool IDictionary<TKey, TValue>.ContainsKey(TKey key)
+    public bool ContainsKey(TKey key)
```
sharwell: :thought_balloon: I find the inconsistency between explicitly and implicitly implemented methods a bit confusing. Note that in addition to understanding why you made the change, I don't have a preference towards implicit or explicit implementations of contract methods overall (each has their merits, and both work).

Microsoft.Research/Contracts/MsCorlib/System.Collections.ObjectModel.Collection.cs
```diff
@@ -0,0 +1,394 @@
...
+
+namespace System.Collections.ObjectModel
+{
+    // Summary:
+    //     Provides the base class for a generic collection.
+    //
+    // Type parameters:
+    //   T:
+    //     The type of elements in the collection.
+    public class Collection<T> : IList<T>, ICollection<T>, IEnumerable<T>, IList, ICollection, IEnumerable
```
sharwell: :bulb: It's difficult to tell while reading this contract class which methods are implementing interface members, and which methods are new for `Collection<T>`. For contracts, this distinction is important because only the latter may define contracts.

Microsoft.Research/Contracts/System/System.Net.Sockets.Socket.cs
tom-englert: Incorrect, many combinations of level and name are valid.

Microsoft.Research/Contracts/MsCorlib/System.Collections.ObjectModel.Collection.cs
```diff
@@ -0,0 +1,394 @@
...
+        //   index:
+        //     The zero-based index at which item should be inserted.
+        //
+        //   item:
+        //     The object to insert. The value can be null for reference types.
+        //
+        // Exceptions:
+        //   System.ArgumentOutOfRangeException:
+        //     index is less than zero.-or-index is greater than System.Collections.ObjectModel.Collection<T>.Count.
+        protected virtual void InsertItem(int index, T item)
```
sharwell: :question: Should this have a contract?

```csharp
Contract.Requires(index >= 0 && index <= Count);
```

Microsoft.Research/Contracts/MsCorlib/System.Math.cs
```diff
@@ -93,29 +93,40 @@ public static Decimal Abs(Decimal value)
     [Pure]
     public static float Abs(float value)
     {
+      Contract.Ensures((Contract.Result<float>() >= 0.0) || float.IsNaN(Contract.Result<float>()) || float.IsPositiveInfinity(Contract.Result<float>()));
+
+      // 2015-03-26: tom-englert
```
sharwell: :exclamation: My notes for the change in contracts for `Rect` apply here as well.

Microsoft.Research/Contracts/MsCorlib/System.Math.cs
```diff
@@ -93,29 +93,40 @@ public static Decimal Abs(Decimal value)
     [Pure]
     public static float Abs(float value)
     {
+      Contract.Ensures((Contract.Result<float>() >= 0.0) || float.IsNaN(Contract.Result<float>()) || float.IsPositiveInfinity(Contract.Result<float>()));
```
sharwell: :question: Since `float.PositiveInfinity >= 0.0f`, why would we need to include it as a separate item in this condition?

Microsoft.Research/Contracts/System.ComponentModel.Composition/Sources/System.ComponentModel.Composition.Hosting.ExportProvider.cs
```diff
@@ -67,12 +67,16 @@ public Lazy<T> GetExport<T>(string contractName)
 
     public T GetExportedValue<T>(string contractName)
     {
+      Contract.Ensures(Contract.Result<T>() != null); 
+      // because of ImportCardinality.ExactlyOne =>
```
sharwell: :bulb: This might be clearer if the comment was placed before the call to `Contract.Ensures`.

Microsoft.Research/Contracts/System.Data/System.Data.DataRowCollection.cs
```diff
@@ -0,0 +1,220 @@
...
+    //}
+
+    /// <summary>
+    /// Gets the row at the specified index.
+    /// </summary>
+    /// 
+    /// <returns>
+    /// The specified <see cref="T:System.Data.DataRow"/>.
+    /// </returns>
+    /// <param name="index">The zero-based index of the row to return. </param><exception cref="T:System.IndexOutOfRangeException">The index value is greater than the number of items in the collection. </exception><filterpriority>1</filterpriority><PermissionSet><IPermission class="System.Security.Permissions.EnvironmentPermission, mscorlib, Version=2.0.3600.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" version="1" Unrestricted="true"/><IPermission class="System.Security.Permissions.FileIOPermission, mscorlib, Version=2.0.3600.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" version="1" PathDiscovery="*AllFiles*"/></PermissionSet>
```
sharwell: :bulb: The formatting here (and in places that follow in the same file) ... "could be improved". :wink:

Microsoft.Research/Contracts/System/System.Collections.ObjectModel.ObservableCollection.cs
```diff
@@ -0,0 +1,193 @@
...
+        //
+        // Parameters:
+        //   oldIndex:
+        //     The zero-based index specifying the location of the item to be moved.
+        //
+        //   newIndex:
+        //     The zero-based index specifying the new location of the item.
+        public void Move(int oldIndex, int newIndex)
+        {
+            Contract.Requires(oldIndex < Count);
```
sharwell: :question: Should this check against 0 as well?

Microsoft.Research/Contracts/System/System.Collections.Specialized.INotifyCollectionChanged.cs
```diff
@@ -0,0 +1,359 @@
...
+        //     class that describes a multi-item change.
+        //
+        // Parameters:
+        //   action:
+        //     The action that caused the event. This can be set to System.Collections.Specialized.NotifyCollectionChangedAction.Reset,
+        //     System.Collections.Specialized.NotifyCollectionChangedAction.Add, or System.Collections.Specialized.NotifyCollectionChangedAction.Remove.
+        //
+        //   changedItems:
+        //     The items that are affected by the change.
+        public NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction action, IList changedItems) { }
```
sharwell: :bulb: Missing contracts?

Microsoft.Research/Contracts/WindowsBase/System.Windows.Threading.DispatcherObject.cs
tom-englert: Done

sharwell:
First of all, this pull request was a **massive** improvement over #21, in terms of ability to focus the review on changes to contracts. Great job with that, I know it must have been a lot of work. :+1: 

You've given me a lot to think about. I'll be filing some "interesting issues" (IMO) as a result. Also :+1: for this.

I'll read over the answers to the questions later today.

Microsoft.Research/Contracts/System/System.Diagnostics.FileVersionInfo.cs
tom-englert: There is a tool "CRASanitizer" that does the verification. When we include the type for `SILVERLIGHT` the build fails.
I just wanted to adopt the existing coding style - I agree things like this should be cleaned up everywhere.

Microsoft.Research/Contracts/MsCorlib/System.Collections.Generic.IDictionary.cs
tom-englert: For a `ContractClass` it's a bit different, it's clear that all methods are interface implementations. Here I changed to implicit methods because it removed unnecessary complexity from some of the contracts.

Microsoft.Research/Contracts/MsCorlib/System.Collections.ObjectModel.Collection.cs
tom-englert: Yes, we should definitely come up with some rules and clean up all the code. Currently I just try to follow the pattern that I find in the current file.

Microsoft.Research/Contracts/MsCorlib/System.Collections.ObjectModel.Collection.cs
tom-englert: Right, I've added this and 2 others.

Microsoft.Research/Contracts/MsCorlib/System.Math.cs
tom-englert: I think it was just to make clear that infinity is allowed here - but maybe also the static checker generated warnings.

Microsoft.Research/Contracts/System.Data/System.Data.DataRowCollection.cs
tom-englert: I just wanted to conform to the existing style :grin:

Microsoft.Research/Contracts/System/System.Collections.ObjectModel.ObservableCollection.cs
tom-englert: Right, I've added this.

Microsoft.Research/Contracts/System/System.Collections.Specialized.INotifyCollectionChanged.cs
tom-englert: Yes, I'll add them later.

sharwell: I'm fine with that as long as you create an issue. If the issue is narrow-scoped, it might make it easier for a new contributor to come along and add them. An example:

> ## Add contracts for NotifyCollectionChangedEventArgs
>
> A stub contract class for `NotifyCollectionChangedEventArgs` was added in #84. The contracts for methods in this class still need to be added.

Then we can mark the issue with the **up for grabs** label. :+1:

Microsoft.Research/Contracts/MsCorlib/System.Collections.ObjectModel.Collection.cs
```diff
@@ -0,0 +1,394 @@
...
+        //   item:
+        //     The new value for the element at the specified index. The value can be null
+        //     for reference types.
+        //
+        // Exceptions:
+        //   System.ArgumentOutOfRangeException:
+        //     index is less than zero.-or-index is greater than System.Collections.ObjectModel.Collection<T>.Count.
+        protected virtual void SetItem(int index, T item)
+        {
+            Contract.Requires((index >= 0) && (index <= Count));
```
sharwell: :exclamation: This is incorrect. It should be `index < Count` instead of `index <= Count` (the documentation above it needs to be corrected as well.

Microsoft.Research/Contracts/System/System.Collections.Specialized.INotifyCollectionChanged.cs
tom-englert: I have added the issue. 
However it would be much easier if the one that finds a bug directly creates the issue :smile:

tom-englert:
@SergeyTeplyakov any idea when you will merge this? @sharwell has reviewed it and I don't see any concerns.

PR #85, opened 2015-07-03
Remove/replace unprovable contracts.

tom-englert:
Some contracts that are theoretically valid but can't be proven by the static checker.
Crated issue #86, so once the checker is improved they can be enabled again.

### DIFF
--- a/Microsoft.Research/Contracts/MsCorlib/System.Math.cs
+++ b/Microsoft.Research/Contracts/MsCorlib/System.Math.cs
@@ -93,29 +93,40 @@ namespace System
     [Pure]
     public static float Abs(float value)
     {
+      Contract.Ensures((Contract.Result<float>() >= 0.0) || float.IsNaN(Contract.Result<float>()) || float.IsPositiveInfinity(Contract.Result<float>()));
+
+      // 2015-03-26: tom-englert
+      // Disabled, since this was too complex for the checker to understand.
+      // e.g. new Rect(0, 0, Math.Abs(a), Math.Abs(b)) raised a warning that with and height are unproven to be positive values.
+
       // !NaN ==>  >= 0
-      Contract.Ensures(float.IsNaN(value) || Contract.Result<float>() >= 0.0);
+      // Contract.Ensures(float.IsNaN(value) || Contract.Result<float>() >= 0.0);
 
       // NaN ==> NaN
-      Contract.Ensures(!float.IsNaN(value) || float.IsNaN(Contract.Result<float>()));
+      // Contract.Ensures(!float.IsNaN(value) || float.IsNaN(Contract.Result<float>()));
 
       // Infty ==> +Infty
-      Contract.Ensures(!float.IsInfinity(value) || float.IsPositiveInfinity(Contract.Result<float>()));
+      // Contract.Ensures(!float.IsInfinity(value) || float.IsPositiveInfinity(Contract.Result<float>()));
 
       return default(float);
     }
 
     [Pure]
     public static double Abs(double value)
-    {      
+    {
+      Contract.Ensures((Contract.Result<double>() >= 0.0) || double.IsNaN(Contract.Result<double>()) || double.IsPositiveInfinity(Contract.Result<double>()));
+
+      // 2015-03-26: tom-englert
+      // Disabled, since this was too complex for the checker to understand.
+      // e.g. new Rect(0, 0, Math.Abs(a), Math.Abs(b)) raised a warning that with and height are unproven to be positive values.
+
       // !NaN ==>  >= 0
-      Contract.Ensures(double.IsNaN(value) || Contract.Result<double>() >= 0.0);
+      // Contract.Ensures(double.IsNaN(value) || Contract.Result<double>() >= 0.0);
 
       // NaN ==> NaN
-      Contract.Ensures(!double.IsNaN(value) || double.IsNaN(Contract.Result<double>()));
-
+      // Contract.Ensures(!double.IsNaN(value) || double.IsNaN(Contract.Result<double>()));
       // Infty ==> +Infty
-      Contract.Ensures(!double.IsInfinity(value) || double.IsPositiveInfinity(Contract.Result<double>()));
+      // Contract.Ensures(!double.IsInfinity(value) || double.IsPositiveInfinity(Contract.Result<double>()));
 
       return default(double);
     }
--- a/Microsoft.Research/Contracts/WindowsBase/System.Windows.Rect.cs
+++ b/Microsoft.Research/Contracts/WindowsBase/System.Windows.Rect.cs
@@ -17,18 +17,6 @@ using System.ComponentModel;
 using System.Diagnostics.Contracts;
 
 
-// 2015-03-36: tom-englert
-// Temporarily disabled checks for Contract.Requires(!this.IsEmpty);
-// => This requirement is true, but it's impossible to proof with acceptable effort.
-//    Even simple code like
-//
-//      var r = new Rect();
-//      r.Widht = 10.0;
-//
-//    will create a warning "CodeContracts: requires unproven: !this.IsEmpty. Are you making some assumption on get_Width that the static checker is unaware of?"
-//    As soon as the checker can infer the proper constraints, this can be enabled again.
-
-
 namespace System.Windows
 {
   // Summary:
--- a/Microsoft.Research/Contracts/WindowsBase/System.Windows.Rect.cs
+++ b/Microsoft.Research/Contracts/WindowsBase/System.Windows.Rect.cs
@@ -16,6 +16,19 @@ using System;
 using System.ComponentModel;
 using System.Diagnostics.Contracts;
 
+
+// 2015-03-36: tom-englert
+// Temporarily disabled checks for Contract.Requires(!this.IsEmpty);
+// => This requirement is true, but it's impossible to proof with acceptable effort.
+//    Even simple code like
+//
+//      var r = new Rect();
+//      r.Widht = 10.0;
+//
+//    will create a warning "CodeContracts: requires unproven: !this.IsEmpty. Are you making some assumption on get_Width that the static checker is unaware of?"
+//    As soon as the checker can infer the proper constraints, this can be enabled again.
+
+
 namespace System.Windows
 {
   // Summary:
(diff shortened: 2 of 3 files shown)


Microsoft.Research/Contracts/WindowsBase/System.Windows.Rect.cs
```diff
@@ -16,6 +16,19 @@
 using System.ComponentModel;
 using System.Diagnostics.Contracts;
 
+
+// 2015-03-36: tom-englert
```
sharwell: :bulb: I prefer these types of longer comment stay in the issue instead of the code. I've found  that over time, code gets corrected but the comments tend to stay forever.

:memo: The other inline comments you added at the location of commented calls to `Contract.Requires` below are different. They are short, and obviously would be changed if someone did implement support for these contracts in the future.

Microsoft.Research/Contracts/WindowsBase/System.Windows.Rect.cs
```diff
@@ -339,7 +352,7 @@ public double Width
       }
       set
       {
-        Contract.Requires(!this.IsEmpty);
+        // Contract.Requires(!this.IsEmpty); => Is true, but impossible to proof with acceptable effort.
```
sharwell: :thought_balloon: It would be really cool if the contract block allowed the use of `Contract.Assume`. For example, this method could use the following contract:

```csharp
Contract.Assume(!this.IsEmpty); // Required, but not currently provable statically
Contract.Requires(value >= 0.0 || Double.IsNaN(value));
```

mike-barnett:
I’m not sure if the best route here is to get rid of the contracts. If they are valid preconditions, but something that the static checker seems unable to prove at call sites (and I’m not sure how you are certain that is true), then you can always add a Contract.Assume just before the call site.

I also don’t understand why you have gotten rid of some postconditions. Why would you do that? The checker never tries to prove a postcondition except when it is verifying the implementation of a method. Are you saying that you are trying to run the static checker on the source of Mscorlib’s implementation?

Mike

tom-englert:
I mainly look at contracts from the users side, not from mathematical theory. 
Contracts are a tool to help writing better code. If it hinders you from writing clean code (e.g. if there are more `Contract.Assume` statements than active code), something is wrong.
This happened to me with Rect and Size, which are heavily used e.g. in layout calculations. Code like
```c#
var r = new Rect();
Contract.Assume(!r.IsEmpty);
r.Width = 5;
Contract.Assume(!r.IsEmpty);
r.Height = 10;
```
is no longer readable. I won't find any acceptance in any team for using CC if people have to write code like this. 

For the post conditions I did not remove them, but simplified them so something that the checker could understand. My reference here was that code like
```c#
var r = new Rect(0, 0, Math.Abs(a), Math.Abs(b))
```
should not generate a warning that width and height are unproven to be positive values. I could not convince anyone that it would be better to write the above statement in 10 lines, where 5 would be just `Contract.Assume`.

It does not make sense building cars if there are no roads. Lets start making some good shoes, and do the cars when we have the roads.

mike-barnett:
Ah, those are great points. Yes, I totally agree in general. But in this case, wouldn't it be better to add a postcondition to the constructor Rect that says it isn't empty? For the latter example, is it that Math.Abs lacks a postcondition or is it that the checker is not good at proving this because they are floating point numbers?

tom-englert:
I have to get some practical sample code to see what was the issue with Size and Rect. 

For Math.Abs I think it is the link between input and output, where it could not decide whether input was some infinity or not, so none of the existing ensures got active.

tom-englert:
The main obstacle here is that Rect and Size are value types, so they can't have a parameterless constructor that could say `Contract.Ensures(!this.IsEmpty)`. If I use e.g. `new Rect(0,0,0,0)` things look better. 
So to fix this we would need 
- a way to add contracts to the default constructor of value types
- someone who checks every method that returns a Rect or a Size and adds the `Contract.Ensures(!Contract.Result<Rect/Size>().IsEmpty)` if applicable. (would be a good job for a trainee)

mike-barnett:
Oh, good point about having contracts on the nullary ctor for structs. Yes, I agree now with your changes to the contracts.

tom-englert:
Created issue #98.

tom-englert:
@mike-barnett any idea why the original contracts of ```Math.Abs``` don't work? 
I'm not sure how the analyzer will handle multiple ```Contract.Ensures```, so I can't tell if they are wrong or just too complex.

On the other hand: Would anyone gain any profit from the original implementation if it would work? Is it worth spending time on this?

brettshearer:
I would expect that if the analyser has problems because the contracts are too complex then there may be benefit in improving the analyser, far outside of the Math.Abs problem.

mike-barnett:
I don't know details of why they don't work, but I know that the checker can have problems with floating point numbers. Multiple postconditions should be just fine: the checker just conjoins them all.

I'm not sure if it is that important at the moment. Maybe just comment out the contracts on Rect (and any similar structs) and leave a comment about why they aren't there.

tom-englert:
OK, that's what the current state is, and we have two issues (#86, #98) to remind us to deal with this later.
@SergeyTeplyakov: So it seems to me this could be merged as well.

Microsoft.Research/Contracts/MsCorlib/System.Math.cs
```diff
@@ -93,29 +93,40 @@ public static Decimal Abs(Decimal value)
...
     {
+      Contract.Ensures((Contract.Result<float>() >= 0.0) || float.IsNaN(Contract.Result<float>()) || float.IsPositiveInfinity(Contract.Result<float>()));
+
+      // 2015-03-26: tom-englert
+      // Disabled, since this was too complex for the checker to understand.
+      // e.g. new Rect(0, 0, Math.Abs(a), Math.Abs(b)) raised a warning that with and height are unproven to be positive values.
+
       // !NaN ==>  >= 0
-      Contract.Ensures(float.IsNaN(value) || Contract.Result<float>() >= 0.0);
+      // Contract.Ensures(float.IsNaN(value) || Contract.Result<float>() >= 0.0);
```
sharwell: :question: What happens if instead of commenting out these `Ensures` lines, just *just* add the new one above?

Microsoft.Research/Contracts/MsCorlib/System.Math.cs
```diff
@@ -93,29 +93,40 @@ public static Decimal Abs(Decimal value)
     [Pure]
     public static float Abs(float value)
     {
+      Contract.Ensures((Contract.Result<float>() >= 0.0) || float.IsNaN(Contract.Result<float>()) || float.IsPositiveInfinity(Contract.Result<float>()));
```
sharwell: :bulb: I would prefer this be written as:

```csharp
Contract.Ensures(float.IsNaN(Contract.Result<float>()) || Contract.Result<float>() >= 0.0);
```

Microsoft.Research/Contracts/MsCorlib/System.Math.cs
```diff
@@ -93,29 +93,40 @@ public static Decimal Abs(Decimal value)
...
+      // Contract.Ensures(!float.IsInfinity(value) || float.IsPositiveInfinity(Contract.Result<float>()));
 
       return default(float);
     }
 
     [Pure]
     public static double Abs(double value)
-    {      
+    {
+      Contract.Ensures((Contract.Result<double>() >= 0.0) || double.IsNaN(Contract.Result<double>()) || double.IsPositiveInfinity(Contract.Result<double>()));
```
sharwell: :bulb: I would prefer this be written as:

```csharp
Contract.Ensures(double.IsNaN(Contract.Result<double>() >= 0.0) || Contract.Result<double>() >= 0.0);
```

Microsoft.Research/Contracts/MsCorlib/System.Math.cs
```diff
@@ -93,29 +93,40 @@ public static Decimal Abs(Decimal value)
     [Pure]
     public static float Abs(float value)
     {
+      Contract.Ensures((Contract.Result<float>() >= 0.0) || float.IsNaN(Contract.Result<float>()) || float.IsPositiveInfinity(Contract.Result<float>()));
+
+      // 2015-03-26: tom-englert
```
sharwell: :bulb: If the `Contract.Ensures` lines below can be preserved in addition to the newly-inserted `Contract.Ensures`, then this comment can be removed. The new `Ensures` line is correct and there was no loss of information to include it.

sharwell:
I understand removing `IsEmpty` from the `Requires` of several methods. I do want to understand two things before merging this though. The first question is added as a diff comment which appears above. The second is:

:question: What happens if instead of commenting out `Contract.Requires(IsEmpty)`, we simply change it to `Contract.Assume(IsEmpty)`?

Microsoft.Research/Contracts/MsCorlib/System.Math.cs
tom-englert: This would work as well, but I found this would be too confusing - if it's commented, it's more clear that it is not active at all.

sharwell: It only shows in the editor extensions quick info if it's active though. :smile: I created #104 to deal with this later since it was clear that you would benefit from having these changes in the first community preview.

Microsoft.Research/Contracts/MsCorlib/System.Math.cs
tom-englert: I think the checker had a problem to infer IsPositiveInfinity from x >= 0; if s.o. requires ```x >= 0 && !IsInfinity(x)```, this could pass if we not explicitly include IsPositiveInfinity!

I will add some test cases once I fully understood the CC regressing testing framework; once we have test cases, we can optimize this and maybe fix the checker to handle double/float more consistently.

Microsoft.Research/Contracts/MsCorlib/System.Math.cs
tom-englert: s.a.

Microsoft.Research/Contracts/MsCorlib/System.Math.cs
tom-englert: s.a.

PR #91, opened 2015-07-08
Fix ccrewriter bugs to support roslyn-based compiler

SergeyTeplyakov:
All tests are passing.

Main difference between old and new compilers are related to "static closures". Here is an example of non-capturing lambda expression (a.k.a. static closure):

```c#
public class Foo
{
     public string Method(params string[] strings)
     {
        Contract.Ensures(Contract.ForAll(strings, s => Contract.Result<string>() == s));
        return "42";
     }
}
```

In this case old compiler will generate following code (pseudo code in C#):

```c#
	public class Foo
	{
		[CompilerGenerated]
		private static Predicate<string> CS$<>9__CachedAnonymousMethodDelegate1;
		public string Method(params string[] strings)
		{
			if (Foo.CS$<>9__CachedAnonymousMethodDelegate1 == null)
			{
				Foo.CS$<>9__CachedAnonymousMethodDelegate1 = new Predicate<string>(Foo.<Method>b__0);
			}
			Contract.Ensures(Contract.ForAll<string>(strings, Foo.CS$<>9__CachedAnonymousMethodDelegate1));
			return "42";
		}
		[CompilerGenerated]
		private static bool <Method>b__0(string s)
		{
			return Contract.Result<string>() == s;
		}
	}
```

But roslyn-based compiler uses different approach. It will generate "static closure class" with instance methods and a singleton instance:

```c#
	public class Foo
	{
		[CompilerGenerated]
		[Serializable]
		private sealed class <>c
		{
			public static readonly Foo.<>c <>9 = new Foo.<>c();
			public static Predicate<string> <>9__0_0;
			internal bool <Method>b__0_0(string s)
			{
				return Contract.Result<string>() == s;
			}
		}
		public string Method(params string[] strings)
		{
			Predicate<string> arg_21_1;
			if ((arg_21_1 = Foo.<>c.<>9__0_0) == null)
			{
				arg_21_1 = (Foo.<>c.<>9__0_0 = new Predicate<string>(Foo.<>c.<>9.<Method>b__0_0));
			}
			Contract.Ensures(Contract.ForAll<string>(strings, arg_21_1));
			return "42";
		}
	}
```

This distinction leads to number of changes to understand new patterns in the IL but preserve backward compatibility.


SergeyTeplyakov:
@mike-barnett, could you please take a look at this PR?

sharwell:
:exclamation: If available, we should pull the Roslyn binaries from NuGet instead of adding them to this pull request.

If the binaries are available on NuGet, we can add a file **.nuget/packages.config** which declares the dependency, [like I did here](https://github.com/tunnelvisionlabs/antlr4cs/blob/master/runtime/CSharp/.nuget/packages.config) for the IKVM dependencies in the ANTLR 4 C# target (which don't apply to any specific project). Then we just update the paths to reference the binaries under the **packages** folder instead of the binaries you included in this pull request.

sharwell:
:memo: I'm going to run a quick test and see if I can send a pull request that changes the current code to use NuGet instead of embedded binaries (no other changes). Then you can update this pull request so commit c1767c5 just updates the version number in **packages.config** and perhaps a path, but otherwise works and is much simpler.

mike-barnett:
@SergeyTeplyakov : Looks great! Let's get a new release out!

sharwell:
:exclamation: So, it turns out the new Roslyn binaries weren't added to this pull request which means two things:

1. It's easier to add them from NuGet (no rebase necessary)
2. The tests fail when run inside Visual Studio on other computers because it can't find the compiler

SergeyTeplyakov:
> :exclamation: So, it turns out the new Roslyn binaries weren't added to this pull request which means two things:

My bad. That's because all executables are excluded in .gitignore. I thought that they were added in previous iterations.

It would be good to add NuGet support. Can you update this PR or what the easiest way to do this?

sharwell:
I'm good with this to get the initial build working.

SergeyTeplyakov:
Thanks, Sam.

PR #46, opened 2015-05-16
Removed invalid preconditions for System.NetSockets.Socket.SetSocketOption(SocketOptionLevel, SocketOptionName, object)

hubuk:
Requirements for SocketOptionLevel and SocketOptionName parameters are too compilated to be included in contracts. At least for now.
Resolves issue #11.

PR #96, opened 2015-07-08
Removed invalid preconditions for Socket.SetSocketOption

### LINKED ISSUE #11: SetSocketOption Contract Requirements are incorrect
For this overload, the contract requires IPV6 (41), but that is not the only valid way of calling this method.

https://github.com/Microsoft/CodeContracts/blob/master/Microsoft.Research/Contracts/System/System.Net.Sockets.Socket.cs#L313

Admittedly, the was this method can be called are not entirely straightforward to express in contract form.

http://msdn.microsoft.com/en-us/library/fs3sawaa%28v=vs.110%29.aspx

sharwell:
The requirements for `SocketOptionLevel` and `SocketOptionName` parameters are too complicated to be included in contracts, at least for now.

Fixes #11
Supersedes #46


sharwell:
I rebased #46 to address the merge conflict. This issue came up previously and I agree with the proposed solution. :+1:

SergeyTeplyakov:
:+1: Thanks a lot @sharwell and @hubuk

PR #97, opened 2015-07-09
[CCCheck] Add an ability to specify db name and parallel connection to the cache

SergeyTeplyakov:
This is a superset of the #80.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -208,3 +208,4 @@ UpgradeLog*.htm
 # Microsoft Fakes
 FakesAssemblies/
 
+Microsoft.Research/CCDoc/Sandcastle/Sandcastle.zip
--- a/Microsoft.Research/Clousot.Cache/ICacheModel.cs
+++ b/Microsoft.Research/Clousot.Cache/ICacheModel.cs
@@ -17,6 +17,7 @@ using System.Collections.Generic;
 using Microsoft.Research.CodeAnalysis.Caching.Models;
 using Microsoft.Research.DataStructures;
 using System.Diagnostics.Contracts;
+using System.Threading.Tasks;
 
 namespace Microsoft.Research.CodeAnalysis.Caching
 {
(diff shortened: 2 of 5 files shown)


SergeyTeplyakov:
Updated the solution.

SergeyTeplyakov:
I thoght a bit more about proposed (and applied) change and I don't think they're good. Original implementation was robust in terms of unobserved exceptions and new one is not. I know that default behavior has been changed and application will not crash by default, but unobserved exceptions should be avoided even if the code would be slightly more complicated. 

I will simplify my original cide and push another update.

sharwell:
> Original implementation was robust in terms of unobserved exceptions and new one is not.

The new implementation is just as robust as the original in terms of unobserved exceptions. Unless we're planning on actually doing something in the event of an exception (previously we weren't), I think we should stick with the code which is much more succinct.

SergeyTeplyakov:
New implementation is not robust:)

Let suppose following situation: this method starts 3 asynchronous operations. First of them will finished first and finished successfully. 

In my original implementation all continuations for all other tasks still would be called and exceptions would be observed. 

In the new implementation, in this case method `CreateCacheAccessorAsync` will return the result as soon as the first task completes. *But* due to `Task.WhenAny` implementation details other running task's exceptions would not be observed!. I.e. if second or third task will fail when the first one will succeeds `Task.UnobservedException` would be raised.

And, as I mentioned for couple of times, the behavior for unobserved tasks has been changed in .NET 4.5 but generally speaking it should be avoided if possible.

sharwell:
> I.e. if second or third task will fail when the first one will succeeds `Task.UnobservedException` would be raised.

Actually, regardless of which one fails you'll get a `Task.UnobservedException`. This is due to the fact that my code checked the task status instead of the `Exception` property.

*But*, this is also the desired behavior because our code is not handling these exceptions. In theory we could log them somewhere, but if we just let them be reported through `UnobservedException` we get more consistent overall behavior (e.g. attach a listener to that event if you want to log the exceptions).

SergeyTeplyakov:
@sharwell I don't think that this is a desired behavior.

And I don't think we should add anything into `CreataAndCheckAsync`. Just because it will violate SRP. Factory method is unaware about high level exception policy, but `CreateCacheAccessorAsync` is aware that we would like to start something in parallel and *ignore* the rest of the world if we connect to first cache. 

The question is: do we really want to switch from one solution that avoids unobserved exceptions to another one that allows them? I intentionally used more complex solution to avoid this issue because for me it is important not to introduce such kind of issues to the codebase.

sharwell:
Following further discussion in Gitter, the summary of my view is the following:

I'm not opposed to avoiding the UnobservedException, but I don't think we should make `CreateCacheAccessorAsync` more complex for that purpose. To avoid the UnobservedException with no other behavior change (because the current behavior is actually correct, it just has the possibility of an undesired unobserved exception), an approach like the [`ObserveExceptions` method](https://github.com/rackerlabs/dotnet-threading/blob/893e19e21c63382792a629b818be0da5f2ce61ef/Rackspace.Threading/CoreTaskExtensions.cs#L1606-L1631) from Rackspace.Threading could be used in the `CreateAndCheckAsync` method to ensure the task returned by `CreateAndCheckAsync` never results in an unobserved exception.

SergeyTeplyakov:
Updated the PR. Added exception observation.

But, as I mentioned in chat: `CreateAndCheckAsync` should not observe potential exceptions from the task it provides. This is a responsibility of the client of this method. So I've added ObservePotentialException method to the `CreateCacheAccessorAsync` that does not complicates the main logic.

sharwell:
:+1:

PR #99, opened 2015-07-09
Features/contract targets

SergeyTeplyakov:
Changed sn.exe resolution based on TargetFrameworkSDKToolsDirectory.

The same approach is used by Roslyn team as well. Example here: https://github.com/dotnet/roslyn/blob/f87d725fb90a713f94732bf035124c3c8df25acb/build/Targets/VSL.Settings.targets#L224


sharwell:
:exclamation: This needs to be rebased onto Microsoft/master. The first four commits listed should not be included.

SergeyTeplyakov:
Working on it.

PR #100, opened 2015-07-09
Changed sn.exe resolution to use TargetFrameworkSDKToolsDirectory

SergeyTeplyakov:
New solution follows a common pattern for sn.exe resolution.

Example could be found in roslyn codebase:
https://github.com/dotnet/roslyn/blob/f87d725fb90a713f94732bf035124c3c8df25acb/build/Targets/VSL.Settings.targets#L224


sharwell:
:+1:

PR #101, opened 2015-07-09
Make sure VS2012 users receive proper update notifications

sharwell:
This commit changes the ID of the Editor Extensions VSIX to match the ID used by the previous [Code Contracts Editor Extensions VS2012](https://visualstudiogallery.msdn.microsoft.com/02de7066-b6ca-42b3-8b3c-2562c7fa024f) extension. By using the same ID, existing users of this extension will receive and upgrade notification after the next release of the editor extensions are made available. Visual Studio 2010 users will not receive this notification even though the extension can be installed in that environment, but that is acceptable since Visual Studio 2010 did not originally have a notification feature.


SergeyTeplyakov:
:+1:

PR #103, opened 2015-07-09
Merge pull request 84

sharwell:
This pull request resolves the conflicts to merge #84.


SergeyTeplyakov:
:+1:

thomas-delahaye:
Unfortunately, the contracts for `IDictionary<TKey, TValue>.ContainsKey` and the like might not be correct for all implementations. The [MSDN site for `IDictionary<TKey, TValue>`](https://msdn.microsoft.com/en-us/library/htszx2dy%28v=vs.110%29.aspx) says in the remarks "Implementations can vary in whether they allow key to be null."...
The same applies to `Add(TKey, TValue)` and `Item[TKey]` (but no remarks for `Remove(TKey)` and `TryGetValue(TKey, out TValue)`...).

sharwell:
@thomas-delahaye Can you create a separate issue for that? This PR is merged and treated as effectively "closed", but I don't think we should lose track of your findings. (and I'm really sorry about the delay in responding!)

PR #105, opened 2015-07-10
Make sure the build updates the Visual Studio 2015 extension version

sharwell:
Previously the build was not updating the extension manifest for the Visual Studio 2015 extension. This meant that the extension manager in Visual Studio 2015 would show the wrong extension version number after an otherwise proper installation.


SergeyTeplyakov:
:+1:

PR #106, opened 2015-07-10
Reformat code in Microsoft.Contracts10

sharwell:
This reformat operation used dotnet/codeformatter@1d719e4e with `/rule-:FieldNames`.

### DIFF
--- a/Foxtrot/Contracts/ContractDeclarativeAssemblyAttribute.cs
+++ b/Foxtrot/Contracts/ContractDeclarativeAssemblyAttribute.cs
@@ -1,16 +1,5 @@
-// CodeContracts
-// 
-// Copyright (c) Microsoft Corporation
-// 
-// All rights reserved. 
-// 
-// MIT License
-// 
-// Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
-// 
-// The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
-// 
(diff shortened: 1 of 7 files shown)


SergeyTeplyakov:
BTW, what do you think about following idea: when I formatted the code I did following changes:

1. Formatted the code using formatter (I've used R#, dotnet/codeformatter would be ok as well)
2. Formatted manually: 
- Added empty line after blocks of code like if statements, foreach statements etc. (R# can't do this automatically, and it seems that dotnet/codeformatter can't do this either).
- Separated logical blocks of code: added blank lines after contract statements, separated assignment statements, put empty line between multiline comments etc.
3. Moved classes into separate files (in many cases files were huge, like 5K lines with 5-10 different classes).
4. Removed regions. There was a tons of places with a regions with a few lines of code.

This approach took much more time than just running the tool but the result was significantly better, IMO. There was a tons of places where code looked like one giant piece of code, without proper line indentation it took me much more time to understand the code.

sharwell:
> * Added empty line after blocks of code like if statements, foreach statements etc.
>
> * Separated logical blocks of code: added blank lines after contract statements, separated assignment statements, put empty line between multiline comments etc.

StyleCopAnalyzers can do some of this automatically for an entire solution, and new rules have been proposed to handle even more of this. I'd like to do it separately from the initial reformat.

The initial reformat covers enough ground that Visual Studio 2013 or 2015 (when configured with the default settings) will not be constantly attempting to reformat the code.

> Moved classes into separate files

StyleCopAnalyzers can't do this *yet*, but we're working on it. I definitely want to automate this so we have a high degree of confidence that the result is correct.

> Removed regions. There was a tons of places with a regions with a few lines of code.

StyleCopAnalyzers can already do this. More importantly, it can distinguish between regions inside and outside of method bodies and only remove the type of regions you select.

SergeyTeplyakov:
Ok. Everything sounds good for me.

sharwell:
:memo: I checked each file in this diff, and it seemed to be correct.

PR #108, opened 2015-07-10
Reformat code in AnalysisTypes

sharwell:
This reformat operation used dotnet/codeformatter@1d719e4e with `/rule-:FieldNames`.


sharwell:
:memo: I checked each file in this diff, and it seemed to be correct.

PR #110, opened 2015-07-10
Updated version info for upcoming release

SergeyTeplyakov:


sharwell:
:exclamation: **VS14.extension.vsixmanifest** needs to be updated the same way as **VS12.extension.vsixmanifest** and the others.

:memo: For future releases, #105 ensures that this is handled automatically. For this release, I'm fine with the change being manually added as a commit here. As long as the version number is consistent, we're good. :+1:

SergeyTeplyakov:
All those changes were made automatically. and, BTW, I've merged #105, so I'll rebuild and will update this PR.

After that I'll create a tag and release page.

sharwell:
:+1:

PR #117, opened 2015-07-11
Update TypeArguments behavior to match old IntelliSense services

### LINKED ISSUE #116: Incorrect usage of TypeArguments property in VS 2015 editor extensions
The Roslyn APIs used for the editor extensions in Visual Studio 2015 have a `TypeArguments` property, but this property does not behave the same way as the similarly-named property in the Visual Studio 2010-2013 IntelliSense services. In particular, for uninstantiated generic types, Visual Studio 2010-2013 treat the type as not having type arguments, but Visual Studio 2015 treats them as having type arguments which are equal to the type parameters of the original definition.

This difference led to an incorrect translation of `ContractsProvider.TryGetTypeReference` in Visual Studio 2015's editor extensions which throws a `StackOverflowException` and crashes the IDE when attempting to display contracts for a member of a generic type.

sharwell:
ISymbolExtensions.TypeArguments needs to return null when the argument is an uninstantiated generic type, or a StackOverflowException can occur when attempting to display contracts for members of a generic type in Visual Studio 2015.

Fixes #116

### DIFF
--- a/Microsoft.Research/ContractAdornments/CSharp.Roslyn/ISymbolExtensions.cs
+++ b/Microsoft.Research/ContractAdornments/CSharp.Roslyn/ISymbolExtensions.cs
@@ -54,6 +54,9 @@ namespace ContractAdornments
             if (namedTypeSymbol == null)
                 return default(ImmutableArray<ITypeSymbol>);
 
+            if (namedTypeSymbol.Equals(namedTypeSymbol.OriginalDefinition))
+                return ImmutableArray<ITypeSymbol>.Empty;
+
             return namedTypeSymbol.TypeArguments;
         }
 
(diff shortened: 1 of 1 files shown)


Microsoft.Research/ContractAdornments/CSharp.Roslyn/ISymbolExtensions.cs
```diff
@@ -54,6 +54,9 @@ public static ImmutableArray<ITypeSymbol> TypeArguments(this ITypeSymbol symbol)
             if (namedTypeSymbol == null)
                 return default(ImmutableArray<ITypeSymbol>);
 
+            if (namedTypeSymbol.Equals(namedTypeSymbol.OriginalDefinition))
```
SergeyTeplyakov: Maybe comment would be useful here? Just to clarify the intent...

sharwell: Done. :smile:

SergeyTeplyakov:
:+1:

PR #121, opened 2015-07-11
Improve identification of relevant nodes for QuickInfo in Roslyn

### LINKED ISSUE #119: QuickInfo does not work for calls in VS 2015 editor extensions
This bug lies in the following block of code in `IntellisenseContractsHelper`:

```csharp
//Is the node in question a node call?
var asCall = nodeInQuestion as InvocationExpressionSyntax;
if (asCall != null) {

  //Make sure we aren't on the right side of the call
  MemberAccessExpressionSyntax memberAccessExpression = asCall.Expression as MemberAccessExpressionSyntax;
  if (memberAccessExpression == null || memberAccessExpression.Name == leafNode)
    return null;

  return asCall;
}
```

The *desired* behavior for the block is to only return the `InvocationExpressionSyntax` if the leaf token being examined is the method name appearing in the invocation expression.

sharwell:
Fixes #119


SergeyTeplyakov:
Merged.

PR #123, opened 2015-07-11
Fix #122

### LINKED ISSUE #122: When toggling between DB versions connecting to the cache fails.
The is a design flaw in the implementation of #80 / #97. If there are both DBs installed, we might end up toggling between both versions if we just use the one that connects first.
If then the file was e.g. generated with v12, and next time v11 connects faster, we end up with "failed to connect to the cache" again!

tom-englert:
Fix #122:
Back to synchronous cache evaluation. Detect proper local db version by looking at the installed versions.


Microsoft.Research/Clousot.Cache/SQLCacheModel.cs
```diff
@@ -784,5 +786,33 @@ protected override SqlConnectionStringBuilder BuildConnectionString(IClousotCach
 
       return base.BuildConnectionString(options);
     }
+
+    public static string GetDesiredDbName()
```
SergeyTeplyakov: Tom, maybe `GetDesiredDbName` should be called outside `LocalDbClousotCacheFactory`? It seems that this is high level responsibility?

Microsoft.Research/Clousot.Cache/SQLCacheModel.cs
```diff
@@ -784,5 +786,33 @@ protected override SqlConnectionStringBuilder BuildConnectionString(IClousotCach
...
+    {
+        var versionKeyFileMapping = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
+        {
+            { @"11.0", @"(LocalDb)\v11.0" },
+            { @"12.0", @"(LocalDb)\MSSQLLocalDB" },
+        };
+
+        using (var rootKey = Registry.LocalMachine.OpenSubKey(@"SOFTWARE\Microsoft\Microsoft SQL Server Local DB\Installed Versions"))
```
SergeyTeplyakov: Could you please extract this path as a const?

Microsoft.Research/Clousot.Cache/SQLCacheModel.cs
```diff
@@ -784,5 +786,33 @@ protected override SqlConnectionStringBuilder BuildConnectionString(IClousotCach
 
       return base.BuildConnectionString(options);
     }
+
+    public static string GetDesiredDbName()
+    {
+        var versionKeyFileMapping = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
```
SergeyTeplyakov: Shoulw we use static field instead? Or even path this as an argument to the GetDesiredDbName?

When I've made this change I tried to separate responsibilities and extracted all "application specific" functionality into high level. That's why I've added dbname as parameter for the constructor.

BTW there is at least one another client of this parameterless constructor and I don't know is this functionality required there or not. Thats why I would suggest to put "desired db name selection" out of this class.

Microsoft.Research/Clousot.Cache/SQLCacheModel.cs
tom-englert: You could (it's public, and there is a constructor that takes the name), but I didn't want to change the calling code too much - now it's just the original code.

Microsoft.Research/ClousotMain/Caching/CacheManager.cs
```diff
@@ -163,65 +162,21 @@ public int NbCacheHits
...
-
-            ObservePotentialExceptions(factoryTasks);
-
-            while (factoryTasks.Count > 0)
-            {
-                Task<IClousotCache> factoryTask = 
-                    await Task.WhenAny(factoryTasks).ConfigureAwait(false);
-
-                if (factoryTask.Status == TaskStatus.RanToCompletion && factoryTask.Result != null)
+                foreach (var factory in cacheAccessorFactories)
```
SergeyTeplyakov: I think even in this case we still would benefit from extracting this logic out of the constructor itself.

Microsoft.Research/Clousot.Cache/SQLCacheModel.cs
SergeyTeplyakov: But as I mentioned, you actually changed the behavior of all clients, but we want to change the behavior for only one of them. Right? 

That's why, if you're concerned about backward compatibility, extracting this logic into the upper level is better.

tom-englert: No, I did not change the behavior of any client (except that it does no longer fail).
There has been always only one instance of ```LocalDbClousotCacheFactory```, and now it will connect to the correct DB. 
We don't need to be backward compatible, the file names are different for every version anyhow.

SergeyTeplyakov: So my point was: extracting this logic our of the factory clearly separates the concerns and responsibilities of the code.

Microsoft.Research/Clousot.Cache/SQLCacheModel.cs
tom-englert: That's why I did it like this - now the other client is no longer broken.
Currently the responsibility of the ```LocalDbClousotCacheFactory``` is to provide a connection to LocalDB - no matter what engine is installed.
If the client ever will need to take responsibility of the database, it can using the proper constructor.

Microsoft.Research/ClousotMain/Caching/CacheManager.cs
tom-englert: Be aware that here is only **one** local db factory and **one** sql db factory with the user configured connection string!

tom-englert: I see all you concerns, but they don't fit in the current strategy of the cache manager, and I would not like to rewrite the cache manager at this point, just fix the issue.

Microsoft.Research/Clousot.Cache/SQLCacheModel.cs
SergeyTeplyakov: BTW, this is IMO, not critical. I'll accept the PR if you don't agree with my comment.

Microsoft.Research/ClousotMain/Caching/CacheManager.cs
SergeyTeplyakov: Ok, so you don't think it is useful to extract this into separate method?

I'm fine with current solution if you think its not worthwhile.

SergeyTeplyakov: > I see all you concerns, but they don't fit in the current strategy of the cache manager, and I would not like to rewrite the cache manager at this point, just fix the issue.

Ok.

SergeyTeplyakov:
:+1: Everything is good.

Microsoft.Research/ClousotMain/Caching/CacheManager.cs
tom-englert: The single responsibility of the ```LocalDbClousotCacheFactory``` is to provide the proper connection string with a valid db name - that's what it does and nothing else. To me that's clean enough.

PR #126, opened 2015-07-11
Reformat code in Graphs

sharwell:
This reformat operation used dotnet/codeformatter@1d719e4e with the default settings.


sharwell:
:memo: I checked all the files in this PR and they seem to be fine.

PR #135, opened 2015-07-13
Fix migration warnings when opening CodeContracts.sln in VS2015

sharwell:
I included a small amount of project cleanup for the VSIX projects in this commit as well.

:memo: The **CodeContractsForRoslyn** project still won't work after this change, but that didn't change.


Microsoft.Research/ContractAdornments/Extension/ContractAdornments.csproj
```diff
@@ -423,13 +423,8 @@
     <CodeAnalysisIgnoreBuiltInRules>false</CodeAnalysisIgnoreBuiltInRules>
     <CodeAnalysisFailOnMissingRules>false</CodeAnalysisFailOnMissingRules>
   </PropertyGroup>
-  <PropertyGroup>
```
sharwell: :memo: The changes to this file should actually have been included in #66 but I forgot this change.

SergeyTeplyakov:
:+1: Looks good for me. Merged.

Issue #142, opened 2015-07-15
Postocondition doesn't work in async method, when using contract class for interface


Dennis-Petrov:
Hi. I'm migrating project from .NET 4 to 4.5.1, using VS 2013, and one of the reasons is to use `async`/`await`. Project widely uses code contracts, most of them are defined for interfaces or abstract classes (hence, there are number of contract classes).

Why this code:

    [ContractClass(typeof(IAContract))]
    interface IA
    {
        Task<int> FooAsync();
    }

    [ContractClassFor(typeof(IA))]
    abstract class IAContract : IA
    {
        // public async Task<int> FooAsync() doesn't help too
        public Task<int> FooAsync()
        {
            Contract.Ensures(Contract.Result<int>() > 0);
            throw new NotImplementedException();
        }
    }

    class A : IA
    {
        public async Task<int> FooAsync()
        {
            await Task.Delay(1000);
            return -1;
        }
    }

doesn't trigger `ContractException`, while this code does:

    class A
    {
        public async Task<int> FooAsync()
        {
            Contract.Ensures(Contract.Result<int>() > 0);
            await Task.Delay(1000);
            return -1;
        }
    }

?

sharwell:
I would actually expect the following code to produce an error when attempting to rewrite the assembly:

    // public async Task<int> FooAsync() doesn't help too
    public Task<int> FooAsync()
    {
        Contract.Ensures(Contract.Result<int>() > 0);
        throw new NotImplementedException();
    }

This code uses `Contract.Result<int>` in a method that returns `Task<int>`. If the method was marked `async` it would be a bit different because you don't have access to the returned `Task`.

Dennis-Petrov:
Well, it doesn't produce an error and successfully compiles. But anyway, if I'll change the signature of contract method to be `async`:

        public async Task<int> FooAsync()
        {
            Contract.Ensures(Contract.Result<int>() > 0);
            throw new NotImplementedException();
        }

or even this way (to avoid warning):

        public async Task<int> FooAsync()
        {
            Contract.Ensures(Contract.Result<int>() > 0);
            await Task.Yield();
            throw new NotImplementedException();
        }

the behavior won't change.

SergeyTeplyakov:
@sharwell Actually, in async method you can reference the task result.

So following code is absolutely correct and lead to contract violation if the result is not 42:

```c#
public static async Task<int> FooAsync()
{
  Contract.Ensures(Contract.Result<int>() == 42);
[...]

PR #144, opened 2015-07-15
Fix for #142. Extractor should not miss inherited preconditions

SergeyTeplyakov:


Foxtrot/Foxtrot/Utils/HelperMethods.cs
```diff
@@ -4186,8 +4186,7 @@ public static Method GetContractMethod(Method instantiatedMethod)
             
             if (candidate != null)
             {
-                if (candidate.Contract != null &&
-                    (candidate.Contract.RequiresCount > 0 || candidate.Contract.EnsuresCount > 0))
+                if (candidate.Contract != null && candidate.Contract.HasAssertions)
```
sharwell: :question: What is `ValidationsCount`? In addition to now checking `AsyncEnsuresCount` here (which I presume fixes the main issue), you now check for `ValidationsCount` as well.

System.Compiler/Nodes.cs
```diff
@@ -16245,6 +16245,20 @@ public RequiresList Validations
       get { return this.validations; }
       set { this.validations = value; }
     }
+
+    /// <summary>
+    /// Returns true if current contract node has at least one assertion in it
+    /// (like pres/posts etc).
```
sharwell: :bulb: It wouldn't hurt to be specific here.

```
/// An assertion is represented by any of the following:
/// <list type="bullet">
/// <item><see cref="O:System.Diagnostics.Contracts.Contract.Requires"/></item>
/// <item><see cref="O:System.Diagnostics.Contracts.Contract.Ensures"/></item>
/// <item>...</item>
/// </list>
```

Foxtrot/Foxtrot/Utils/HelperMethods.cs
SergeyTeplyakov: Actually I've extracted this code from another place, not from this one. There is another place that needs to be consistent.

There is a plenty of precondition types. `ValidationCount` is a number of method calls abbreviated with `ContractValidationAttribute`. All of them should be considered in this case.

System.Compiler/Nodes.cs
SergeyTeplyakov: Assertion is a very common term in this domain. I can't understand why to duplicate the info here.

SergeyTeplyakov: I meant in DbC.

SergeyTeplyakov: Ok will change to preconditions/postconditions.

SergeyTeplyakov: Yep, actually this would be more accurate statement. Because if a method contains another assertion like `Contract.Assert`/`Contract.Assume` this method will not return true.

sharwell:
Assuming the tests pass, LGTM. :+1:

Foxtrot/Foxtrot/Utils/HelperMethods.cs
sharwell: Sounds like you fixed another bug (and one which might be much older) in addition to the original issue which deals only with async methods.

SergeyTeplyakov:
Yep, tests are passing:)

And I'll share some day how to work on rewriter: how to add tests, how to debug them etc.

Issue #148, opened 2015-07-18
Fix broken regression tests


tom-englert:
Some of the regression tests that were running fine with the initial check in now fail. 
Some reasons I could identify are
- system contracts have been updated, so there are sometimes different/new messages
- the messages are the same, but the IL offsets have shifted (maybe due to rewriter fixes)

Regression tests are located in folder Regressions\Clousot```, projects ```ClousotTests``` and ```ClousotCacheTests```. All tests are named ```Analyze*``` or ```Slice*```

sharwell:
@tom-englert The big trouble I had is it was unclear which tests were expected to pass and which tests were expected to fail (baseline). I would love to have a command line argument which builds the test projects and runs them (with the console runner) using the expected test settings for each assembly, so we can get a baseline.

**buildCC.cmd** doesn't build the tests. We'll need another build command for testing that uses (I believe) configuration `Debug` and platform `Mixed Platforms`.

tom-englert:
@sharwell I think every test passed with the original code. I could make lots of them run with very few changes, which were all reasonable. 
I did not try everything yet, because the worst problem with these tests is that one test takes about **half an hour** to run, and there are lots of them.

SergeyTeplyakov:
All ccrewrite tests are fine, the only remaining part is cccheck tests. And they could be broken because of last changes and @tom-englert, you're absolutely right, it is very hard to run them because they take too long.

tom-englert:
I could prove that Analyze1FromSourcesV40 was running with the baseline, and managed to have it succeed with the latest code. 
However when looking deeper into the other tests I found this a great mess of stale code and lots of unclear things.
e.g. it runs tests on 
- clousot2.exe
- clousot2s.exe
- Clousot2_Queue.exe
- Cloudot.exe

Are these needed or are they only experimental? Will anything else but clousot.exe be deployed?
@mike-barnett maybe we can get some statements of the team members that did this code what was really needed, or what is only experimental and could be purged. Also I could get only a rough overview of the regression test framework - some basic documentation on this would be very helpful.

Issue #149, opened 2015-07-18
Text formatting does not specify culture


tom-englert:
String formatting in the analyzer does not specify a culture, therefore some tests fail on systems with non-English culture with e.g. "expected: 2.0, actual: 2,0"

I have added a workaround in ```ClousotMain.Analyze()``` by setting the main threads culture to invariant. However this is very bad practice, and should be fix in long term.

SergeyTeplyakov:
Yep, totally agree. It should be automated way to switch to culture agnostic formatting. I think R# could do this.

tom-englert:
Currently it's hard to use R# because the formatting of the code is not yet cleaned up. 
Once we have the code properly formatted, we could add a R#.sln.DotSettings file so everyone is using the same style.

SergeyTeplyakov:
I meant that even in this case it is possible to run only one fix across all projects. I'll try to do this.

sharwell:
> Once we have the code properly formatted, we could add a R#.sln.DotSettings file so everyone is using the same style.

I'm not using ReSharper. I'm planning to add a proposal that we use StyleCopAnalyzers for style checking. It has no dependencies other than Roslyn, is free for everyone, is extremely fast, and we've implemented the ability to fix most violations with 3 clicks or less.

tom-englert:
I don't mind using StyleCopAnalyzers, I think it does not contradict with making the live of R# users easier :smile:

PR #151, opened 2015-07-19
Fix regression tests (Analyze1FromSourcesV40) 

tom-englert:
Fix the Analyze1FromSourcesV40 regression tests. (for #148)
- offset changes
- changes due to fixed system contracts

Should not be pulled before #152, #153 and #154

### DIFF
--- a/Microsoft.Research/ClousotMain/ClousotMain.cs
+++ b/Microsoft.Research/ClousotMain/ClousotMain.cs
@@ -36,6 +36,7 @@ using System.Diagnostics;
 using Microsoft.Research.ClousotPulse.Messages;
 using Microsoft.Research.CodeAnalysis.Inference;
 using System.IO.Pipes;
+using System.Globalization;
 
 // Because these classes and methods can be called from different places (Console, Visual Studio, WCF Service, ...)
 // - do not use Console unless guarded with #if DEBUG. Instead use output
@@ -795,6 +796,10 @@ namespace Microsoft.Research.CodeAnalysis
 
       public int Analyze()
       {
+        // String formatting in the analyzer does not specify a culture, therefore some tests fail on systems with non-English culture with e.g. "expected: 2.0, actual: 2,0"
+        // Setting the threads culture should be removed as soon as issue #149 is fixed.
+        Thread.CurrentThread.CurrentCulture = CultureInfo.InvariantCulture;
+
         try
         {
           return InternalAnalyze();
--- a/Microsoft.Research/RegressionTest/ClousotTests/ClousotTests.csproj
+++ b/Microsoft.Research/RegressionTest/ClousotTests/ClousotTests.csproj
@@ -179,6 +179,12 @@
     <None Include="..\TestFrameworkOOB\UserFeedback.cs">
       <Link>Sources\UserFeedback.cs</Link>
     </None>
+    <Compile Include="..\IteratorAnalysis\IteratorSimpleContract\IteratorSimpleContract.cs">
+      <Link>Sources\IteratorSimpleContract.cs</Link>
+    </Compile>
+    <Compile Include="..\TestPostconditionInference\Filtering.cs">
+      <Link>Sources\Filtering.cs</Link>
+    </Compile>
     <Compile Include="Tests.cs" />
     <None Include="..\Containers\TestContainers\ArrayWithNonNullAnalysis.cs">
       <Link>Sources\ArrayWithNonNullAnalysis.cs</Link>
--- a/Microsoft.Research/RegressionTest/ClousotTests/Sources/Abbreviators.cs
+++ b/Microsoft.Research/RegressionTest/ClousotTests/Sources/Abbreviators.cs
@@ -63,7 +63,7 @@ namespace ClousotTests
         }
 
         [ClousotRegressionTest]
-        [RegressionOutcome(Outcome = ProofOutcome.False, Message = "ensures is false: this.X == Contract.OldValue(this.X)", PrimaryILOffset = 19, MethodILOffset = 20)]
+        [RegressionOutcome(Outcome = ProofOutcome.False, Message = "ensures is false: X == Contract.OldValue(X)", PrimaryILOffset = 19, MethodILOffset = 20)]
         [RegressionOutcome(Outcome = ProofOutcome.Bottom, Message = @"ensures unreachable", PrimaryILOffset = 43, MethodILOffset = 20)]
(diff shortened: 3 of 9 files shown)


SergeyTeplyakov:
I'm going to merge this stuff. Any objections? @tom-englert, @sharwell

sharwell:
Merging #152, #153, and #154 should not be a problem. After that, we will need to rebase this pull request before it can be merged.

tom-englert:
I will rebase this as soon as the others are merged.

SergeyTeplyakov:
I assume that the fix was rebased. If there is no objections I'll merge the PR.

sharwell:
@SergeyTeplyakov No, this pull request is still waiting for #152, #153, and #154 to be merged.

SergeyTeplyakov:
Merged #152, #153, and #154.

sharwell:
:+1:

tom-englert:
Rebased, ready to pull :+1:

Microsoft.Research/RegressionTest/ClousotCacheTests/Properties/AssemblyInfo.cs
```diff
@@ -1,5 +1,16 @@
-// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+﻿// CodeContracts
```
sharwell: :exclamation: Please revert this header change (and any like it).

Microsoft.Research/ClousotMain/ClousotMain.cs
```diff
@@ -795,6 +796,10 @@ private bool FilterPreconditionsAccordingToWarningLevel(BoxedExpression exp)
 
       public int Analyze()
       {
+        // String formatting in the analyzer does not specify a culture, therefore some tests fail on systems with non-English culture with e.g. "expected: 2.0, actual: 2,0"
+        // Setting the threads culture should be removed as soon as issue #149 is fixed.
+        Thread.CurrentThread.CurrentCulture = CultureInfo.InvariantCulture;
```
sharwell: :exclamation: I would be comfortable with this in the test suite, but not in the application itself. We should find another way if at all possible.

sharwell:
:exclamation: Can you rebase this to not include the first three commits?

Microsoft.Research/ClousotMain/ClousotMain.cs
tom-englert: The other way is to fix thousands of formatting statements where the culture is missing. 

Pros for this way of fixing (until all the code is culture-specific):
- All the code has never been verified using another culture than US(~Invariant). 
- The only thing affected by this would be the UI - but the analyzer does not have one.

sharwell: Can we not put it in the test suite initialization code?

tom-englert: The tests are just calling the .exe - we would need another argument to pass the culture to the other process.

Microsoft.Research/RegressionTest/ClousotCacheTests/Properties/AssemblyInfo.cs
tom-englert: So the two lines header is the correct one?

Microsoft.Research/ClousotMain/ClousotMain.cs
sharwell: > The tests are just calling the .exe - we would need another argument to pass the culture to the other process.

That would be preferable. My vote is for [`/LCID` to match **devenv.exe**](https://msdn.microsoft.com/en-us/library/9cytz106.aspx).

tom-englert:
OK, just committed the functional changes. 
I thought the formatting changes would be gone after merging automatically, but somehow they survived.

sharwell:
Changes to fix the tests look fine. I would prefer to separate the culture change to its own pull request, but I'll leave the call up to @SergeyTeplyakov. Overall :+1:.

tom-englert:
p.s. there is already an issue to make string-formatting culture-independent: #149

Microsoft.Research/RegressionTest/ClousotTests/ClousotTests.csproj
```diff
@@ -179,6 +179,12 @@
     <None Include="..\TestFrameworkOOB\UserFeedback.cs">
       <Link>Sources\UserFeedback.cs</Link>
     </None>
+    <Compile Include="..\IteratorAnalysis\IteratorSimpleContract\IteratorSimpleContract.cs">
+      <Link>Sources\IteratorSimpleContract.cs</Link>
+    </Compile>
+    <Compile Include="..\TestPostconditionInference\Filtering.cs">
+      <Link>Sources\Filtering.cs</Link>
```
panacekcz: This file causes ClousotTests not to show in Test Explorer, because it contains methods declared `extern`.

    Error loading C:\…\CodeContracts\Microsoft.Research\RegressionTest\ClousotTests\bin\Debug\ClousotTests.dll: Unable to load the test container 'C:\…\CodeContracts\Microsoft.Research\RegressionTest\ClousotTests\bin\Debug\ClousotTests.dll' or one of its dependencies. Error details:
    System.TypeLoadException: Could not load type 'TestPostconditionInference.FilterRedundantPostcondition' from assembly 'ClousotTests, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' because the method 'GetApplicationVersion' has no implementation (no RVA).
    System.TypeLoadException: Could not load type 'TestPostconditionInference.Array' from assembly 'ClousotTests, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' because the method 'get_Length' has no implementation (no RVA).

Microsoft.Research/ClousotMain/ClousotMain.cs
panacekcz: :bulb: What about using invariant culture if `/regression` option is used? Possible implementation: 4c98ec5a4e181a9d05b627f04c56507317d0c3ec
I think a more complicated `/LCID` does not add much value (why would you use it other than for this?).

Microsoft.Research/RegressionTest/ClousotTests/ClousotTests.csproj
SergeyTeplyakov: @sharwell I prefer separate PR, but separate commits are fine for now.

@tom-englert What do you think about @panacekcz comment? We can fix this in separate commit/PR.

tom-englert: These files should have been added as <None>, not as <Compile>. 
It's only to have the test files part of the project.

PR #152, opened 2015-07-19
Reformat code in ClousotTests

sharwell:
This reformat operation used dotnet/codeformatter@1d719e4e with `/rule-:FieldNames`. In addition, two other changes were made:

1. Use **Tests.cs** instead of **v.cs** (the former is exactly equivalent to the latter aside from including one additional disabled test).
2. Add the remaining test sources to the project.


Microsoft.Research/RegressionTest/ClousotTests/AsyncTestDriver.cs
```diff
@@ -20,147 +9,146 @@
...
-    // We have no control on the order of the tests, so we make sure
-    // to always call Begin before End
-
-    public void BeginTest(Options options)
-    {
-      if (this.skipTest(options))
-        return;
+        public AsyncTestDriver(Action<Options, Output> action, Func<Options, bool> skipTest, uint maxWaitHandles)
+        {
+            this.action = action;
```
SergeyTeplyakov: I think I saw that in some other PRs with formatting changes you also changed field names...

Microsoft.Research/RegressionTest/ClousotTests/AsyncTestDriver.cs
```diff
@@ -20,147 +9,146 @@
...
 
-    public void EndTest(Options options)
-    {
-      if (this.skipTest(options))
-        return;
+            beginTestsProcessed = true;
 
-      if (!this.beginTestsProcessed)
-        this.orderReversed = true;
+            if (orderReversed)
```
SergeyTeplyakov: Can formatter insert braces for single statements like that?

Microsoft.Research/RegressionTest/ClousotTests/AsyncTestDriver.cs
sharwell: I explicitly disabled that - I left a message in Gitter explaining why (basically there is so much code to reformat that we need to do the safe option first and then we can consider renaming things in the future).

I closed all the other formatting PRs so I can update them to also *not* rename fields.

Microsoft.Research/RegressionTest/ClousotTests/AsyncTestDriver.cs
sharwell: This is following the .NET foundation bracing rules, which generally requires braces but makes an exception for cases where the entire block is a single statement which does not span multiple lines. This bracing rule was later implemented in StyleCopAnalyzers as well (DotNetAnalyzers/StyleCopAnalyzers#716, DotNetAnalyzers/StyleCopAnalyzers#717).

Microsoft.Research/RegressionTest/ClousotTests/AsyncTestDriver.cs
SergeyTeplyakov: Sounds good. Jut wanted to double check.

Microsoft.Research/RegressionTest/ClousotTests/AsyncTestDriver.cs
SergeyTeplyakov: There is a plenty of places in this code base like:

```c#
if (check)
{
  // several statements
}
else
   oneStatement;
```

This is very unreadable. I would prefer to not use braces only on a simple if statements (like old-style preconditions) but only when there is no `else` blocks. Is this consistent with .NET foundation bracing rules?

sharwell: No, if the `if` statement requires braces then the `else` statement must also use braces (and vice-versa). This will eventually be enforced by DotNetAnalyzers/StyleCopAnalyzers#716 (which provides document-, project-, and solution-wide automatic fixes), whether or not codeformatter.exe handles it in this pull request.

SergeyTeplyakov: It seems I didn't communicate my thoughts correctly. I'll try once more.

My rules: use curly braces in all cases. The only exception is simple if-statements without else blocks for short one liners like argument validation.

sharwell: StyleCopAnalyzers and StyleCop do not currently have a rule to enforce this, though one could certainly be proposed. However, there are a couple additional things to keep in mind:

1. I'm very much hoping that the Code Contracts projects can be moved from the Microsoft organization to the dotnet organization. If it does move here, it would make sense to use the formatting rules of other projects in that organization even if they deviate somewhat from our personal preferences.
1. The initial formatting is being done with dotnet/codeformatter (both in support of the first item and because it is compatible with Visual Studio's default formatting settings) so strict brace inclusion rules might be best saved for a later time.

SergeyTeplyakov: Actually, I'm just curios how far my personal preferences from proposed once:) And yep, moving to dotnet organization makes a lot of sense. I'll talk to someone to understand how to do this.

PR #158, opened 2015-07-20
xunit preview

sharwell:
# DO NOT MERGE (not now, not ever)

**This pull request exists only as a preview to demonstrate potential improvements to the testing process by using features from xUnit.net.**

@tom-englert You will definitely want to review this potential solution and help determine if we want to go "all-in" on the approach.

Currently the tests take a long time to run because each "unit test" in the clousot and foxtrot test suites is actually a suite of tests driven by an XML file. xUnit.net provides a similar feature via its `MemberData` attribute. However, unlike MSTest, this approach results in each test appearing individually in Test Explorer, which means they can be *run* individually in Test Explorer.

xUnit.net may also allow us to parallelize some of the tests, but I have not yet determined whether or not this is the case.

### DIFF
--- a/Microsoft.Research/RegressionTest/ClousotTests/v.cs
+++ b/Microsoft.Research/RegressionTest/ClousotTests/v.cs
@@ -12,8 +12,10 @@
 // 
 // THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+using System.Collections.Generic;
 using ClousotTests;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Xunit;
 
 namespace Tests
 {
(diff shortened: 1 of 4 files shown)


Microsoft.Research/RegressionTest/ClousotTests/v.cs
```diff
@@ -48,6 +50,1516 @@ public TestContext TestContext
       }
     }
 
+    private static readonly Options[] TestOptions =
+    {
+        new Options(
```
sharwell: :memo: Each `new Options` here is drawn from an XML element from the original test inputs file.

Microsoft.Research/RegressionTest/ClousotTests/v.cs
```diff
@@ -48,6 +50,1516 @@ public TestContext TestContext
...
+            sourceFile: @"Microsoft.Research\RegressionTest\ClousotTests\Sources\IOperations.cs",
+            compilerOptions: "",
+            libPaths: new string[0],
+            references: new[] { "" },
+            options: "-infer autopropertiesensures -suggest requires -bounds -show:validations;unreached",
+            contractReferenceAssemblies: false,
+            exe: false,
+            compiler: "CS"),
+#if false
+  new Options(
```
sharwell: :memo: I did some simple find/replace below this marker, but they haven't been completely converted yet. Now that I have a better idea what the output should look like, I'll do the final conversion using an XSLT.

Microsoft.Research/RegressionTest/ClousotTests/v.cs
```diff
@@ -48,6 +50,1516 @@ public TestContext TestContext
...
+    options: "-infer autopropertiesensures -suggest requires -includesuggestionsinregression -premode combined -suggest requires -nonnull -bounds -show unreached -show progress -check assumptions -infer methodensures -suggest asserttocontracts"
+    contractReferenceAssemblies: true"
+    exe: false"
+     ),
+  new Options(
+    sourceFile: @"Microsoft.Research\RegressionTest\Inference\AssertToContracts.cs"
+    compilerOptions: "/optimize /define:AGGRESSIVEINFERENCE"
+    libPaths: new string[0],
+    references: new[] { "System.Core.dll"
+    options: "-infer autopropertiesensures -suggest requires -includesuggestionsinregression -premode combined -suggest requires -nonnull
```
sharwell: :memo: It didn't work to return the `Options` instances directly from this member data property. It may have something to do with the way test discovery and executions run in separate AppDomains. Rather than deal with the underlying cause, I found that simply returning and index was a very easy solution that is sure to work.

Microsoft.Research/RegressionTest/ClousotTests/v.cs
```diff
@@ -216,6 +1728,18 @@ public void Slice2SequentialFromSourcesV40()
         TestDriver.BuildAndAnalyze2Slicing(options);
     }
 
+    [Theory]
+    [MemberData("ClousotTestInputs")]
+    public void XunitSlice2SequentialFromSourcesV40(int optionsIndex)
```
sharwell: :memo: We create one `[Theory]` method from each of the existing test methods that references **ClousotTestInputs.xml**, and xUnit.net will automatically break up each theory into 175 (currently) discreet unit tests.

tom-englert:
@sharwell Sounds great, the current solution is a nightmare regarding maintenance and performance.
But first we should cleanup things - there are lot of stale files not used in any test, and tests for stuff that seems to be experimental only. Do you have any idea what 
    clousot2.exe
    clousot2s.exe
    Clousot2_Queue.exe
    Cloudot.exe
are used for?

mike-barnett:
Clousot2 is a version of the static checker that uses CCI2 (i.e., the CCI that is in CodePlex) as its binary reader. Clousot (the project that ends up being cccheck.exe) uses an older internal version of CCI (we called it CCI1) that is located in the System.Compiler folder. Cloudot is a cloud-based version of Clousot that (I believe) was meant to parallelize the static analysis.

tom-englert:
So cloudot seems to be a dead end, but what about clousot2 and its derivates?

mike-barnett:
Well, CCI2 is much better than CCI1, so I would recommend moving to using that. But CCI2 is now not really supported either so maybe it doesn't matter. There were some subtle differences in the regressions when using CCI2 instead of CCI1, but we never got a chance to finish looking into that before Francesco left.

Roslyn will, of course, supersede all of this. But I don't know when: they currently do not have any provision for reading binaries.

Issue #157, opened 2015-07-20
Capturing async postcondition generates invalid IL


SergeyTeplyakov:
Yet another issue with async postconditions (see #155, #156 for other cases).

If async postcondition captures state, resulting IL is invalid. Here is an example:

```c#
    class Foo
    {
        private int _limit = 42;
        private bool _wasThrown = false;
        public async Task<int> FooAsync()
        {
            // Currently this code lead to unverifiable code!
            Contract.Ensures(Contract.Result<int>() == _limit);
            Contract.EnsuresOnThrow<InvalidOperationException>(_wasThrown);

            await Task.Delay(42);

            return 42;
        }
    }
```

In this case ccrewrite will generate following code for async closure `CheckPost` method:

```c#
private class <FooAsync>AsyncContractClosure_0
{
	public Task<int> CheckPost(Task<int> task)
	{
		TaskStatus status = task.Status;
		if (status == TaskStatus.RanToCompletion && __ContractsRuntime.insideContractEvaluation <= 4)
		{
			try
			{
				__ContractsRuntime.insideContractEvaluation++;
				Task<int> task2;
				RewriterMethods.Ensures(task2.Result == this._limit, null, "Contract.Result<int>() == _limit");
			}
			finally
			{
				__ContractsRuntime.insideContractEvaluation--;
			}
		}
		TaskStatus status2 = task.Status;
		if (status2 == TaskStatus.Faulted)
		{
			AggregateException exception = task.Exception;
			bool flag = this.CheckException(exception);
		}
		return task;
	}
	public bool CheckException(Exception e)
	{
		InvalidOperationException ex = e as InvalidOperationException;
		if (ex != null)
		{
			RewriterMethods.EnsuresOnThrow(this._wasThrown, null, "_wasThrown", ex);
		}
		else
		{
			AggregateException ex2 = e as AggregateException;
			if (ex2 != null)
			{
				Func<Exception, bool> predicate = new Func<Exception, bool>(this.CheckException);
				ex2.Handle(predicate);
			}
		}
		return true;
	}
}
```

So both methods: `CheckPost` and `CheckException` are using `this._limit` and `this._wasThrown` respectively. But those fields are in the parent class, there is now such fields in the closure class.

This code will not fail during runtime, but peverify will complain about type mismatch.

UPDATE: changing `Foo` from `class` to `struct` will lead to runtime failure: `InvalidProgramException`.

sharwell:
[...]

Issue #156, opened 2015-07-20
Async postconditions are broken when task is cancelled


SergeyTeplyakov:
As I mentioned in #155, async postconditions are implemented incorrectly right now.

Here is two additional examples.

```c#
class Foo
{
    private static bool _exceptionWasThrown;
    private static async Task FooAsync(CancellationToken token)
    {
        Contract.EnsuresOnThrow<InvalidOperationException>(_exceptionWasThrown);

        await Task.Delay(4200, token);

        //return 42;
    }

    public static async Task HandleFooAsync()
    {
        try
        {
            var cts = new CancellationTokenSource();
            cts.Cancel();
            // If async postcondition are implemented properly
            // they should not affect exception handling when the method throws.
            await FooAsync(cts.Token);

            throw new Exception("This should never happen!!");
        }
        catch (TaskCanceledException)
        {
            Console.WriteLine("Task was cancelled!");
        }
        catch (Exception e)
        {
            Console.WriteLine("UNEXPECTED: " + e);
        }
    }
}
```

To deal with `EnsuresOnThrow` ccrewrite generates `bool CheckException(Exception)` method that will throw if the assertion specified in a `Contract.EnsuresOnThrow` is not held. This method implemented correctly, but the call-site is wrong.

ccrewrite generates two different code patterns for generic and non-generic tasks and both of them lead to undesired behavior.

In the previous example non-generic tas was used, so ccrewrite will generate following code:

```c#
	public void CheckPost(Task task)
	{
		AggregateException ex = task.Exception as AggregateException;
		if (ex != null)
		{
			Func<Exception, bool> predicate = new Func<Exception, bool>(this.CheckException);
			ex.Handle(predicate);
			throw ex;
		}
	}
```

ccrewrite also mde following changes to the original async method:

```c#
private static Task FooAsync(CancellationToken token)
{
	Program.Foo.<FooAsync>AsyncContractClosure_0 @object = new Program.Foo.<FooAsync>AsyncContractClosure_0();
	Task task;
	try
	{
		Program.Foo.<FooAsync>d__9 <FooAsync>d__;
		<FooAsync>d__.token = token;
		<FooAsync>d__.<>t__builder = AsyncTaskMethodBuilder.Create();
		<FooAsync>d__.<>1__state = -1;
		AsyncTaskMethodBuilder <>t__builder = <FooAsync>d__.<>t__builder;
[...]

Issue #155, opened 2015-07-20
Async postconditions wraps original exception into additional AE


SergeyTeplyakov:
Current implementation of the asynchronous postconditions is wrong. Consider following code:

```c#
static async Task<int> WillThrow()
{
    Contract.Ensures(Contract.Result<int>() == 42);

    await Task.Delay(42);
    throw new InvalidOperationException();
}

static async Task Run()
{
    try
    {
        await WillThrow();
    }
    catch (InvalidOperationException)
    {
        Console.WriteLine("Expected!");
    }
    catch (AggregateException ae)
    {
        Console.WriteLine("NOT EXPECTED!");
    }
}
```

In this case `Expected!` should be printed with or without any postconditions in the `WillThrow` method. Unfortunately, right now `NOT EXPECTED!` would be printed.

Here is a reason for this. ccrewrite uses following approach: `AsyncClosure` class is generated with `int CheckPost(Task<int> task)` method:

```c#
public int CheckPost(Task<int> task)
{
		int result = task.Result;
		if (__ContractsRuntime.insideContractEvaluation <= 4)
		{
			try
			{
				__ContractsRuntime.insideContractEvaluation++;
				__ContractsRuntime.Ensures(task.Result == 42, null, "Contract.Result<int>() == 42");
			}
			finally
			{
				__ContractsRuntime.insideContractEvaluation--;
			}
		}
		return result;
}
```

And following transformation are made for original async method:

```c#
public Task<int> WillThrow(string str)
{
	ThrowsWithAsyncPostconditionMixed.<WillThrow>AsyncContractClosure_0 @object = new ThrowsWithAsyncPostconditionMixed.<WillThrow>AsyncContractClosure_0();
	__ContractsRuntime.Requires(str != null, null, "str != null");
	ThrowsWithAsyncPostconditionMixed.<WillThrow>d__0 <WillThrow>d__;
	<WillThrow>d__.<>4__this = this;
	<WillThrow>d__.str = str;
	<WillThrow>d__.<>t__builder = AsyncTaskMethodBuilder<int>.Create();
	<WillThrow>d__.<>1__state = -1;
	AsyncTaskMethodBuilder<int> <>t__builder = <WillThrow>d__.<>t__builder;
	<>t__builder.Start<ThrowsWithAsyncPostconditionMixed.<WillThrow>d__0>(ref <WillThrow>d__);
	Task<int> task = <WillThrow>d__.<>t__builder.Task;
[...]

PR #159, opened 2015-07-20
Reimplemented async closure generation

SergeyTeplyakov:
This is a fist part of changes that addresses numerous issues related to async postconditions.

This PR addresses #155 and #156, but not #157 yet.

# Rationale

As I mentioned in #155 and #156 current implementation of the async closure is not correct and changes behavior and semantic of the production code.

New approach is much more simple and does not changes behavior of the code.

# New approach

`EmitAsyncClosure` class generate `AsyncClosure` with `CheckMethod` and `CheckException` methods.

Following transformation is applied to the original async method:

```c#
// This is original task, generated by the AsyncTaskMethodBuilder
var originalTask = t_builder.Task;

var closure = new AsyncClosure();
var task2 = originalTask.ContinueWith(closure.CheckPost).Unwrap();
return task2;
```

To use this approach `CheckPost` method's return type should be changed from `void`/`int` to `Task`/`Task<int>`.

There are 2 cases:

1) Task has no return value. In this case only `EnsuresOnThrow` could be used, and we emit:

```c#
Task CheckMethod(Task t)
{
  if (t.Status == TaskStatus.Faulted)
  {
    // CheckException will throw if EnsuresOnThrow is not held
    CheckException(t.Exception);
  }

  return t;
}
```

2) Task(T) reutrns a T value. In this case both `EnsuresOnThrow` and `Contract.Ensures(Contract.Result)` could be used.

```c#
Task<int> CheckMethod(Task<int> t)
{
  if (t.Status == TaskStatus.Faulted)
  {
    // CheckException will throw if EnsuresOnThrow is not held
    CheckException(t.Exception);
  }
  
  if (t.Status == TaskStatus.RanToCompletion)
  {
     // Check ensures
  }
}
```

This approach changes the behavior of the async method only when postcondition is violated but not changes the behavior otherwise.


sharwell:
New approach looks good. I haven't looked at the code yet. I don't see anything about `TaskContinuationOptions.ExecuteSynchronously`; we'll definitely want to make sure we're passing that to `ContinueWith` or we'll be facing a huge performance impact from the rewriter.

SergeyTeplyakov:
I would like to separate those fixes. I'll create separate work item for it and will fix using separate PR.

Foxtrot/Driver/Rewriting/Utils/TypeNodeExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+    internal static class TypeNodeExtensions
+    {
+        public static int TemplateArgumentsCount(this TypeNode typeNode)
+        {
+            Contract.Requires(typeNode != null);
+
+            return typeNode.TemplateArguments == null ? 0 : typeNode.TemplateArguments.Count;
+        }
+
+        public static int CountOrDefault(this TypeNodeList list)
```
sharwell: :thought_balloon: I don't like this extension method. Soon we will be able to use the following, and until then I think we should just write the expression where it's used.

```
list?.Count();
```

Foxtrot/Driver/Rewriting/Utils/TypeNodeExtensions.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Compiler;
+using System.Diagnostics.Contracts;
+
+namespace Microsoft.Contracts.Foxtrot.Utils
+{
+    internal static class TypeNodeExtensions
+    {
+        public static int TemplateArgumentsCount(this TypeNode typeNode)
```
sharwell: :thought_balloon: I don't like this extension method for the same reason as `CountOrDefault`. The way to write this is the following:

```
typeNode.TemplateArguments?.Count ?? 0;
```

Foxtrot/Driver/Rewriting/Utils/ParameterExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Microsoft.Contracts.Foxtrot.Utils
+{
+    internal static class ParameterExtensions
+    {
+        /// <summary>
+        /// Returns underlying parameter type.
+        /// For generic types it would be like "Func`2".
+        /// </summary>
+        public static string GetGenericTypeName(this Parameter parameter)
```
sharwell: :bulb: I would rather this be an extension method on `TypeNode`, and more accurately reflect the behavior:

```csharp
/// <summary>
/// Gets the metadata name of the type. For generic types the arity is appended, such
/// as <c>Func`2</c> for <see cref="Func{T, TResult}"/>.
/// </summary>
public static string GetMetadataName(this TypeNode typeNode)
{
  Contract.Requires(typeNode != null);
  return HelperMethods.Unspecialize(typeNode).Name.Name;
}
```

Foxtrot/Driver/Rewriting/ReplaceResult.cs
```diff
@@ -163,7 +163,8 @@ public ReplaceResult(Method containingMethod, Local r, Module assemblyBeingRewri
         ///        return "42";
         ///    }
         ///
-        /// In this case, caller of this code can decide to emit a warning, because currently resultin 
+        /// In this case, caller of this code can decide to emit a warning, because current behaviour
```
sharwell: :bulb: This is a US-based project, so the expected spelling here is 'behavior'.

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
```diff
@@ -0,0 +1,969 @@
...
+
+            Method contractEnsuresMethod = this.rewriter.RuntimeContracts.EnsuresMethod;
+
+            foreach (Ensures e in GetTaskResultBasedEnsures(asyncPostconditions))
+            {
+                // TODO: Not sure that 'break' is enough! It seems that this is possible
+                // only when something is broken, because normal postconditions 
+                // are using Contract.Result<T>() and this is possible only for
+                // generic tasks.
+                if (IsVoidTask()) break; // something is wrong in the original contract
```
sharwell: :question: Should this condition instead throw?

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
```diff
@@ -0,0 +1,969 @@
...
+
+            // Wrapping normal ensures into following if statement
+            // if (task.Status == TaskStatus.RanToCompletion)
+            // { postcondition check }
+            //
+            // Implementation of this stuff is a bit tricky because if-statements
+            // are inverse in the IL.
+            // Basically, we need to generate following code:
+            // if (!(task.Status == Task.Status.RanToCompletion))
+            //   goto EndOfNormalPreconditions;
```
sharwell: :bulb: You use 'precondition' many times from here through where exceptional postconditions section starts. Do you mean to use 'postcondition' instead?

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
```diff
@@ -0,0 +1,969 @@
...
+                            new MethodCall(new MemberBinding(null, 
+                                checkExceptionMethod),
+                            new ExpressionList(checkExceptionMethod.ThisParameter, aeLocal))));
+
+                checkExceptionBlock.Statements.Add(endOfExceptionPrecondition);
+                
+                methodBody.Add(checkExceptionBlock);
+            }
+
+            // Previously, following line was executed only without exceptional postconditions!
```
sharwell: :exclamation: If "previously" does not mean "earlier in this method", it should be removed. Maintain focus instead of the current behavior and the reason it is correct.

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
```diff
@@ -0,0 +1,969 @@
...
+            // no exceptional post conditions
+            newBodyBlock.Statements.Add(origBody);
+
+            Block returnBlock = CreateReturnBlock(checkMethodTaskParameter, lastEnsuresSourceContext);
+            methodBody.Add(returnBlock);
+        }
+
+        private static IEnumerable<Ensures> GetTaskResultBasedEnsures(List<Ensures> asyncPostconditions)
+        {
+            foreach (var post in asyncPostconditions)
```
sharwell: :bulb: This entire block is just:

```csharp
return asyncPostConditions.Where(post => !(post is EnsuresExceptional));
```

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
```diff
@@ -0,0 +1,969 @@
...
+                if (!(post is EnsuresExceptional))
+                {
+                    yield return post;
+                }
+            }
+        }
+
+        private static IEnumerable<EnsuresExceptional> GetExceptionalEnsures(List<Ensures> asyncPostconditions)
+        {
+            foreach (var post in asyncPostconditions)
```
sharwell: :bulb: This entire block is just:

```csharp
return asyncPostConditions.OfType<EnsuresExceptional>();
```

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
```diff
@@ -0,0 +1,969 @@
...
+        /// </summary>
+        [Pure]
+        private static Member GetUnwrapMethod(TypeNode checkMethodTaskType)
+        {
+            Contract.Requires(checkMethodTaskType != null);
+            Contract.Ensures(Contract.Result<Member>() != null);
+
+            // TODO: use static for this?
+            var taskExtensions =
+                    HelperMethods.FindType(SystemCoreAssembly,
```
sharwell: :question: Will this still work for corefx, where `TaskExtensions` is not in System.Core? I don't think we need to address it now, but if we know it will fail for some cases that users will want it probably makes sense to add an issue to track it.

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
```diff
@@ -0,0 +1,969 @@
...
+            // We emit the following method:
+            //   bool CheckException(Exception e) {
+            //     var ex = e as C1;
+            //     if (ex != null) {
+            //       EnsuresOnThrow(predicate)
+            //     }
+            //     else {
+            //       var ex2 = e as AggregateException;
+            //       if (ex2 != null) {
+            //         ex2.Handle(CheckException);
```
sharwell: :exclamation: This seems incorrect. Async/await only unwrap exactly the first exception, and only unwrap one level. I'm willing to create a separate issue to track this.

Foxtrot/Driver/Rewriting/Utils/TypeNodeExtensions.cs
SergeyTeplyakov: I now that this solution is far from perfect AND it would be better to use new language feature one day. But I disagree that this solution brings no value.

Compare two different versions:

```c#
                    var templateArgCount = (originalInstanceMethod.TemplateArguments == null)
                        ? 0
                        : originalInstanceMethod.TemplateArguments.Count;

                    for (int i = 0; i < templateArgCount; i++)
                    {
                        args.Add(originalInstanceMethod.TemplateArguments[i]);
                    }
```

and

```c#
                    for (int i = 0; i < originalInstanceMethod.TemplateArguments.CountOrDefault(); i++)
                    {
                        args.Add(originalInstanceMethod.TemplateArguments[i]);
                    }
```

And please note that there is a few places like that in the code. 

I'm OK to add TODO to change this after migration to C# 6.0, but *I'm not OK* to remove to old version.

Foxtrot/Driver/Rewriting/Utils/TypeNodeExtensions.cs
SergeyTeplyakov: The same as before. But this time I prefer this extension method over your suggested solution. 

Extension method is much more descriptive and high level. Proposed solution required "slow memory" to analyze, IMO.

Foxtrot/Driver/Rewriting/Utils/ParameterExtensions.cs
SergeyTeplyakov: Sounds good. :+1:

Foxtrot/Driver/Rewriting/ReplaceResult.cs
SergeyTeplyakov: :+1:

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
SergeyTeplyakov: This one was old behavior and I don't want to change existing behavior to much. I've faced a few issues when my obviously innocent change break existing code. The problem with fail-fast behavior for such a change that it could break existing customers.

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
SergeyTeplyakov: Thanks. This stuff is only for async *postconditions*. Will fix this stuff. :+1:

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
SergeyTeplyakov: Ok.

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
SergeyTeplyakov: I thought about it. I think proper exception if the type is missing is required.

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
SergeyTeplyakov: Please do not forget that async postconditions could be applied to *ANY* task-based methods! But even for simple async methods it is possible that another method was called that throws `AggregateException` with much more complex exception tree inside.

So I do think it is correct. 

P.S. This is an old code, I didn't touch it.

Foxtrot/Driver/Rewriting/Utils/TypeNodeExtensions.cs
sharwell: :arrow_right: Let's create an issue for it. Initially mark it as **blocked** on updating to development in VS 2015, but also mark it with an **easy** label to draw in new contributors.

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
sharwell: :arrow_right: I'm not sure this code or associated comment(s) could be improved within the scope of this pull request, so I'm :ok: with it at least for now.

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
sharwell: :arrow_right: Assuming this code is covered by the tests, it will also be covered if/when we start to test against CoreCLR/DNX/other newness. I think getting a test to cover this (which you already have) is the important part for this PR.

SergeyTeplyakov: I can add additional test case that will run the Foxtrot.exe without System.Core and will expect this new exception type.

sharwell:
@SergeyTeplyakov Once you update this PR to include the items you indicated you plan to change, I think we're good on this one. :+1:

SergeyTeplyakov:
@sharwell Thanks a lot for thoughtful review! Will push an update tomorrow.

sharwell:
The following is a comment in #157 about the impact:

> It is not Roslyn-specific. Async postconditions were never implemented properly.

:arrow_right: Based on this comment, I don't think we need to wait for this pull request to fix #157 before it can be merged.

SergeyTeplyakov:
Yep. I'll merge this pr later today.

PR #166, opened 2015-07-22
Fix warnings showing in the wrong pane in Visual Studio 2015

### LINKED ISSUE #137: Warnings and messages from static contract checking don't appear in the VS2015 error list
I've been testing the last two releases in VS2015 RC. I've noticed that while warnings and messages from the static analysis are appearing in the Output window, they do not show up in the Error List window. The messages do still appear in the Error List in VS2013, so this seems to be specific to VS2015.

This is easy to repro. Just create a new console app and add the class below. Turn on static contract checking in the project properties, then build. In the output window, you'll see the following warning, but it won't show up in VS2015's error list (it will if you do the same thing in 2013).

Projects\ConsoleApplication2\Program.cs(20,4): warning : CodeContracts: Missing precondition in an externally visible method. Consider adding Contract.Requires(list != null); for parameter validation

	public class Foo
	{
		public void DoStuff(List<int> list)
		{
			Console.WriteLine(list.Count);
		}
	}

sharwell:
* Fix #137 (easy)
* Fix the build to actually include the results of these changes in the new installers (not easy :expressionless: )


Microsoft.VisualStudio.CodeTools/PropertyPage/Common.cs
```diff
@@ -52,7 +52,7 @@ static public String GetLocalRegistryRoot()
             // Get the visual studio root key
             if (VsRoot == null)
             {
-                ILocalRegistry2 localReg = GetService(typeof(ILocalRegistry)) as ILocalRegistry2;
+                ILocalRegistry2 localReg = GetService(typeof(SLocalRegistry)) as ILocalRegistry2;
```
SergeyTeplyakov: Just curious what was the reason for this switch?

SergeyTeplyakov:
Looks good to me!

Microsoft.VisualStudio.CodeTools/PropertyPage/Common.cs
sharwell: From the [documentation for `ILocalRegistry2`](https://msdn.microsoft.com/en-us/library/microsoft.visualstudio.shell.interop.ilocalregistry2.aspx):

> **Notes to Callers**
> Called by any package that wants to register in the Visual Studio section of the registry. This interface is obtained by using the [`SLocalRegistry`](https://msdn.microsoft.com/en-us/library/microsoft.visualstudio.shell.interop.slocalregistry.aspx) service.

buildCC.bat
```diff
@@ -1,4 +1,10 @@
 @echo off
+
+cd Microsoft.VisualStudio.CodeTools\CodeToolsSetup
+call buildMSM release
+call export release
```
sharwell: :memo: It would be better if the return codes from these calls were checked.

Microsoft.VisualStudio.CodeTools/PropertyPageUI/PropertyPageUI.rc
```diff
@@ -44,7 +46,29 @@ BEGIN
     "#define VER_FILEDESCRIPTION_STR ""PropertyPage Package Resources\\0""\r\n"
     "#define VER_INTERNALNAME_STR ""PropertyPageUI.dll\\0""\r\n"
     "#define VER_ORIGINALFILENAME_STR ""PropertyPageUI.dll\\0""\r\n"
-    "#include ""EnvSdkVersion.ver""\r\n"
+    "VS_VERSION_INFO VERSIONINFO\r\n"
```
sharwell: :memo: The other option is creating the **EnvSdkVersion.ver** file with these contents.

Issue #170, opened 2015-07-25
Missing contracts in System.Linq.Expressions.Expression and derived types


fedotovalex:
The current version of the contract assembly for System.Core.dll misses contracts for most of the elements added in .NET 4:
- most static methods in Expression (e.g. Expression.AddAssign or new overloads of Expression.Call)
- Update methods in specific expression classes (e.g. BinaryExpression.Update)
- methods and properties in new classes added in .NET 4 (e.g. GotoExpression)

fedotovalex:
Fixed by #171.

PR #171, opened 2015-07-25
Add missing contracts in System.Linq.Expressions.Expression and derived types

fedotovalex:
This addresses #170. Adds contracts for: 
- static methods in the Expression class
- Update methods in specific expression classes
- methods and properties in new classes added in .NET 4

### DIFF
--- a/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.DynamicExpression.cs
+++ b/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.DynamicExpression.cs
@@ -15,11 +15,13 @@
 #if NETFRAMEWORK_4_0 || SILVERLIGHT_4_0 || SILVERLIGHT_5_0
 // File System.Linq.Expressions.DynamicExpression.cs
 // Automatically generated contract file.
-using System.Collections.Generic;
-using System.IO;
-using System.Text;
-using System.Diagnostics.Contracts;
 using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Diagnostics.Contracts;
+using System.IO;
+using System.Runtime.CompilerServices;
+using System.Text;
 
 // Disable the "this variable is not used" warning as every field would imply it.
 #pragma warning disable 0414
@@ -49,34 +51,39 @@ namespace System.Linq.Expressions
       return default(Expression);
     }
 
+    [Pure]
     public DynamicExpression Update (IEnumerable<Expression> arguments)
     {
+      Contract.Ensures(Contract.Result<DynamicExpression>() != null);
       return default(DynamicExpression);
     }
     #endregion
 
     #region Properties and indexers
-    public System.Collections.ObjectModel.ReadOnlyCollection<Expression> Arguments
+    public ReadOnlyCollection<Expression> Arguments
     {
       get
       {
-        return default(System.Collections.ObjectModel.ReadOnlyCollection<Expression>);
+        Contract.Ensures(Contract.Result<ReadOnlyCollection<Expression>>() != null);
+        return default(ReadOnlyCollection<Expression>);
       }
     }
 
-    public System.Runtime.CompilerServices.CallSiteBinder Binder
+    public CallSiteBinder Binder
     {
       get
       {
-        return default(System.Runtime.CompilerServices.CallSiteBinder);
+        Contract.Ensures(Contract.Result<CallSiteBinder>() != null);
+        return default(CallSiteBinder);
       }
     }
 
-    public System.Type DelegateType
+    public Type DelegateType
     {
       get
       {
-        return default(System.Type);
+        Contract.Ensures(Contract.Result<Type>() != null);
+        return default(Type);
       }
     }
 
--- a/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.GotoExpression.cs
+++ b/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.GotoExpression.cs
@@ -72,6 +72,7 @@ namespace System.Linq.Expressions
     public GotoExpression Update (LabelTarget target, Expression value)
     {
       Contract.Requires(target != null);
+      Contract.Requires(value != null || target.Type == typeof(void), "Label type must be void when no value is provided");
       Contract.Ensures(Contract.Result<GotoExpression>() != null);
       return default(GotoExpression);
     }
--- a/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.GotoExpression.cs
+++ b/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.GotoExpression.cs
@@ -68,8 +68,11 @@ namespace System.Linq.Expressions
       return default(Expression);
     }
 
+    [Pure]
     public GotoExpression Update (LabelTarget target, Expression value)
     {
+      Contract.Requires(target != null);
+      Contract.Ensures(Contract.Result<GotoExpression>() != null);
       return default(GotoExpression);
     }
     #endregion
@@ -95,6 +98,7 @@ namespace System.Linq.Expressions
     {
       get
       {
+        Contract.Ensures(Contract.Result<LabelTarget>() != null);
         return default(LabelTarget);
       }
     }
--- a/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.IndexExpression.cs
+++ b/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.IndexExpression.cs
@@ -15,11 +15,12 @@
 #if NETFRAMEWORK_4_0 || SILVERLIGHT_4_0 || SILVERLIGHT_5_0
 // File System.Linq.Expressions.IndexExpression.cs
 // Automatically generated contract file.
+using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Diagnostics.Contracts;
 using System.IO;
 using System.Text;
-using System.Diagnostics.Contracts;
-using System;
 
 // Disable the "this variable is not used" warning as every field would imply it.
 #pragma warning disable 0414
@@ -49,18 +50,21 @@ namespace System.Linq.Expressions
       return default(Expression);
     }
 
+    [Pure]
     public IndexExpression Update (Expression @object, IEnumerable<Expression> arguments)
     {
+      Contract.Ensures(Contract.Result<IndexExpression>() != null);
       return default(IndexExpression);
     }
     #endregion
 
     #region Properties and indexers
-    public System.Collections.ObjectModel.ReadOnlyCollection<Expression> Arguments
+    public ReadOnlyCollection<Expression> Arguments
     {
       get
       {
-        return default(System.Collections.ObjectModel.ReadOnlyCollection<Expression>);
+        Contract.Ensures(Contract.Result<ReadOnlyCollection<Expression>>() != null);
+        return default(ReadOnlyCollection<Expression>);
       }
     }
 
--- a/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.ListInitExpression.cs
+++ b/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.ListInitExpression.cs
@@ -37,6 +37,7 @@ namespace System.Linq.Expressions
       get
       {
         Contract.Ensures(Contract.Result<ReadOnlyCollection<ElementInit>>() != null);
+        Contract.Ensures(Contract.Result<ReadOnlyCollection<ElementInit>>().Count >= 1);
         return default(ReadOnlyCollection<ElementInit>);
       }
     }
--- a/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.ListInitExpression.cs
+++ b/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.ListInitExpression.cs
@@ -13,9 +13,10 @@
 // THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 using System;
+using System.Collections.Generic;
 using System.Collections.ObjectModel;
-using System.Text;
 using System.Diagnostics.Contracts;
+using System.Text;
 
 namespace System.Linq.Expressions
 {
@@ -55,5 +56,16 @@ namespace System.Linq.Expressions
         return default(NewExpression);
       }
     }
+
+#if NETFRAMEWORK_4_0 || SILVERLIGHT_4_0 || SILVERLIGHT_5_0
+    [Pure]
+    public ListInitExpression Update(NewExpression newExpression, IEnumerable<ElementInit> initializers)
+    {
+      Contract.Requires(newExpression != null);
+      Contract.Requires(initializers != null);
+      Contract.Ensures(Contract.Result<ListInitExpression>() != null);
+      return default(ListInitExpression);
+    }
+#endif
   }
 }
--- a/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.ListInitExpression.cs
+++ b/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.ListInitExpression.cs
@@ -58,6 +58,22 @@ namespace System.Linq.Expressions
     }
 
 #if NETFRAMEWORK_4_0 || SILVERLIGHT_4_0 || SILVERLIGHT_5_0
+    //
+    // Summary:
+    //     Creates a new expression that is like this one, but using the supplied children.
+    //     If all of the children are the same, it will return this expression.
+    //
+    // Parameters:
+    //   newExpression:
+    //     The System.Linq.Expressions.ListInitExpression.NewExpression property of the
+    //     result.
+    //
+    //   initializers:
+    //     The System.Linq.Expressions.ListInitExpression.Initializers property of the result.
+    //
+    // Returns:
+    //     This expression if no children are changed or an expression with the updated
+    //     children.
     [Pure]
     public ListInitExpression Update(NewExpression newExpression, IEnumerable<ElementInit> initializers)
     {
--- a/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.LoopExpression.cs
+++ b/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.LoopExpression.cs
@@ -53,6 +53,7 @@ namespace System.Linq.Expressions
     public LoopExpression Update (LabelTarget breakLabel, LabelTarget continueLabel, Expression body)
     {
       Contract.Requires(body != null);
+      Contract.Requires(continueLabel == null || continueLabel.Type == typeof(void), "Continue label type must be void");
       Contract.Ensures(Contract.Result<LoopExpression>() != null);
       return default(LoopExpression);
     }
@@ -80,6 +81,7 @@ namespace System.Linq.Expressions
     {
       get
       {
+        Contract.Ensures(Contract.Result<LabelTarget>() == null || Contract.Result<LabelTarget>().Type == typeof(void));
         return default(LabelTarget);
       }
     }
--- a/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.LoopExpression.cs
+++ b/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.LoopExpression.cs
@@ -49,8 +49,11 @@ namespace System.Linq.Expressions
       return default(Expression);
     }
 
+    [Pure]
     public LoopExpression Update (LabelTarget breakLabel, LabelTarget continueLabel, Expression body)
     {
+      Contract.Requires(body != null);
+      Contract.Ensures(Contract.Result<LoopExpression>() != null);
       return default(LoopExpression);
     }
     #endregion
@@ -60,6 +63,7 @@ namespace System.Linq.Expressions
     {
       get
       {
+        Contract.Ensures(Contract.Result<Expression>() != null);
         return default(Expression);
       }
     }
--- a/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.MemberExpression.cs
+++ b/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.MemberExpression.cs
@@ -50,6 +50,7 @@ namespace System.Linq.Expressions
       get
       {
         Contract.Ensures(Contract.Result<MemberInfo>() != null);
+        Contract.Ensures(Contract.Result<MemberInfo>() is FieldInfo || Contract.Result<MemberInfo>() is PropertyInfo);
         return default(MemberInfo);
       }
     }
--- a/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.MemberExpression.cs
+++ b/Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.MemberExpression.cs
@@ -53,5 +53,26 @@ namespace System.Linq.Expressions
         return default(MemberInfo);
       }
     }
+
+#if NETFRAMEWORK_4_0 || SILVERLIGHT_4_0 || SILVERLIGHT_5_0
+    //
+    // Summary:
+    //     Creates a new expression that is like this one, but using the supplied children.
+    //     If all of the children are the same, it will return this expression.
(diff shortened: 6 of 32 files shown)


sharwell:
@fedotovalex This is great! :smile: Note that I'm hand reviewing every added contract for accuracy, so please **do not** rebase, squash, or amend any commits in this pull request (doing so will force me to start over). I have not found any mistakes yet, but even if I do please just correct them as new commits.

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.DynamicExpression.cs
```diff
@@ -49,34 +51,39 @@ protected internal override Expression Accept (ExpressionVisitor visitor)
       return default(Expression);
     }
 
+    [Pure]
     public DynamicExpression Update (IEnumerable<Expression> arguments)
     {
```
SergeyTeplyakov: Does it means that arguments could be null?

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.DynamicExpression.cs
```diff
@@ -49,34 +51,39 @@ protected internal override Expression Accept (ExpressionVisitor visitor)
...
+      Contract.Ensures(Contract.Result<DynamicExpression>() != null);
```
sharwell: :bulb: The current implementation does not require arguments to be non-null, but I'm not sure that was the intended behavior. It's not intuitive or mentioned in the documentation, and isn't consistent with other `Update` methods. Consider adding a requirement that `arguments` not be null.

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.IndexExpression.cs
```diff
@@ -49,18 +50,21 @@ protected internal override Expression Accept (ExpressionVisitor visitor)
       return default(Expression);
     }
 
+    [Pure]
     public IndexExpression Update (Expression @object, IEnumerable<Expression> arguments)
     {
+      Contract.Ensures(Contract.Result<IndexExpression>() != null);
```
sharwell: :thought_balloon: Interesting. The code backs this up, though I've never seen a static indexer (and don't think you can create or use them in C#).

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.DynamicExpression.cs
SergeyTeplyakov: Yep, it seems that `argument` could be null. This is a valid case.

sharwell: :thought_balloon: Just because the implementation doesn't crash when `arguments` is null doesn't mean the contract needs to allow null.

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.DynamicExpression.cs
fedotovalex: Yeah, this was a great source of confusion to me. It appears that as a general rule they promote nulls to empty collections, unless that collection is expected to be non-empty. If that's the case they require it be non-null.

There are many methods in Expression that don't require non-null on IEnumerable<> and params arrays for that reason. Should those be changed as well?

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.GotoExpression.cs
```diff
@@ -68,8 +68,11 @@ protected internal override Expression Accept (ExpressionVisitor visitor)
       return default(Expression);
     }
 
+    [Pure]
     public GotoExpression Update (LabelTarget target, Expression value)
     {
+      Contract.Requires(target != null);
+      Contract.Ensures(Contract.Result<GotoExpression>() != null);
```
SergeyTeplyakov: Following requires could be added:

`Contract.Requies(value != null || target.Type == typeof(void), "Label must be void or have an Expression");`

From http://referencesource.microsoft.com/#System.Core/Microsoft/Scripting/Ast/GotoExpression.cs,357

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.DynamicExpression.cs
sharwell: > There are many methods in Expression that don't require non-null on `IEnumerable<>` and params arrays for that reason. Should those be changed as well?

I'd like to discuss this after I finish reviewing it, and we can make a decision on whether to leave them all as-is or change them all.

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.ListInitExpression.cs
```diff
@@ -55,5 +56,32 @@ public NewExpression NewExpression
...
+    //
+    // Returns:
+    //     This expression if no children are changed or an expression with the updated
+    //     children.
+    [Pure]
+    public ListInitExpression Update(NewExpression newExpression, IEnumerable<ElementInit> initializers)
+    {
+      Contract.Requires(newExpression != null);
+      Contract.Requires(initializers != null);
+      Contract.Ensures(Contract.Result<ListInitExpression>() != null);
```
SergeyTeplyakov: I see another precondition here: http://referencesource.microsoft.com/#System.Core/Microsoft/Scripting/Ast/ListInitExpression.cs,231

`Contract.Requires(initializers.Any());`

Not sure that this common, but it seems that those methods are not possible to use with infinite sequences.

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.MemberExpression.cs
```diff
@@ -1,57 +1,78 @@
...
+    //   expression:
+    //     The System.Linq.Expressions.MemberExpression.Expression property of the result.
+    //
+    // Returns:
+    //     This expression if no children are changed or an expression with the updated
+    //     children.
+    [Pure]
+    public MemberExpression Update(Expression expression)
+    {
+      Contract.Ensures(Contract.Result<MemberExpression>() != null);
```
sharwell: :question: Should we add something like this?

```
Contract.Requires(Member.IsStatic || expression != null);
Contract.Requires(!Member.IsStatic || expression == null);
```

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.LoopExpression.cs
```diff
@@ -49,8 +49,11 @@ protected internal override Expression Accept (ExpressionVisitor visitor)
       return default(Expression);
     }
 
+    [Pure]
     public LoopExpression Update (LabelTarget breakLabel, LabelTarget continueLabel, Expression body)
     {
+      Contract.Requires(body != null);
+      Contract.Ensures(Contract.Result<LoopExpression>() != null);
```
SergeyTeplyakov: From this line of code: http://referencesource.microsoft.com/#System.Core/Microsoft/Scripting/Ast/LoopExpression.cs,137

Additional requies could be added: `Contract.Requies(continueLabel == null || continueLabel.Type != typeof(void), "Label type must be void");`

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.DynamicExpression.cs
SergeyTeplyakov: @sharwell I don't think we can and should add this precondition. You may find this code: http://referencesource.microsoft.com/#System.Core/Microsoft/Scripting/Ast/DynamicExpression.cs,710

This code means that `null` is an absolutely valid input.

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.SwitchCase.cs
```diff
@@ -44,8 +45,11 @@ sealed public partial class SwitchCase
     #region Methods and constructors
     private SwitchCase() {}
 
+    [Pure]
     public SwitchCase Update (IEnumerable<Expression> testValues, Expression body)
     {
+      Contract.Requires(body != null);
```
sharwell: :bulb: You can definitely require `testValues` not be null here (enforced by the implementation).

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.IndexExpression.cs
SergeyTeplyakov: I think you can't create them in C#, but you can create them in VB.NET.

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.SwitchExpression.cs
```diff
@@ -49,18 +50,22 @@ protected internal override Expression Accept (ExpressionVisitor visitor)
       return default(Expression);
     }
 
+    [Pure]
     public SwitchExpression Update (Expression switchValue, IEnumerable<SwitchCase> cases, Expression defaultBody)
     {
+      Contract.Requires(switchValue != null);
```
sharwell: :bulb: The implementation requires `cases` not be null or empty.

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.SwitchCase.cs
SergeyTeplyakov: @sharwell again. We can't add this as a precondition. `null` is an absolutely valid value. Please look here: http://referencesource.microsoft.com/#System.Core/Microsoft/Scripting/Ast/SwitchCase.cs,100

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.SwitchExpression.cs
SergeyTeplyakov: And again, this is not the case: http://referencesource.microsoft.com/#System.Core/Microsoft/Scripting/Ast/SwitchExpression.cs,204

Any enumerable that is used in a form of `someEnumerable.ToReadOnly()` could be null.

SergeyTeplyakov: Ok, it seems proper precondition should be like that:

`Contract.Requires(cases != null && cases.Any());`
`Contract.Requires(Contract.ForAll(cases, case => case != null));`

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.MemberExpression.cs
SergeyTeplyakov: Theoretically, we can add another precondition:

`Contract.Requires(Member is FieldInfo || Member is PropertyInfo, "Member should be field or property");

Based on this: http://referencesource.microsoft.com/#System.Core/Microsoft/Scripting/Ast/MemberExpression.cs,365

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.MethodCallExpression.cs
```diff
@@ -63,5 +64,29 @@ public MethodInfo Method
...
+    //   arguments:
+    //     The System.Linq.Expressions.MethodCallExpression.Arguments property of the result.
+    //
+    // Returns:
+    //     This expression if no children are changed or an expression with the updated
+    //     children.
+    [Pure]
+    public MethodCallExpression Update(Expression @object, IEnumerable<Expression> arguments)
+    {
+      Contract.Ensures(Contract.Result<MethodCallExpression>() != null);
```
SergeyTeplyakov: The same is true here (as @sharwell suggested earlier):

```c#
Contract.Requires(Member.IsStatic || expression != null);
Contract.Requires(!Member.IsStatic || expression == null);
```

And from this: http://referencesource.microsoft.com/#System.Core/Microsoft/Scripting/Ast/BinaryExpression.cs,95e70d48bc8fe361

We can add:

`Contract.Requires(!Method.IsGenericMethodDefinition);`
`Contract.Requires(!Method.ContainsGenericParameters);`

:question: What do you think?

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.NewArrayExpression.cs
```diff
@@ -44,5 +45,27 @@ public ReadOnlyCollection<Expression> Expressions
...
+    //     The System.Linq.Expressions.NewArrayExpression.Expressions property of the result.
+    //
+    // Returns:
+    //     This expression if no children are changed or an expression with the updated
+    //     children.
+    [Pure]
+    public NewArrayExpression Update(IEnumerable<Expression> expressions)
+    {
+      Contract.Requires(expressions != null);
+      Contract.Ensures(Contract.Result<NewArrayExpression>() != null);
```
SergeyTeplyakov: `Contract.Requires(Type.GetElementType() != typeof(void))`

From this: http://referencesource.microsoft.com/#System.Core/Microsoft/Scripting/Ast/NewArrayExpression.cs,150

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.NewExpression.cs
```diff
@@ -71,5 +72,26 @@ public ReadOnlyCollection<MemberInfo> Members
...
+    //   arguments:
+    //     The System.Linq.Expressions.NewExpression.Arguments property of the result.
+    //
+    // Returns:
+    //     This expression if no children are changed or an expression with the updated
+    //     children.
+    [Pure]
+    public NewExpression Update(IEnumerable<Expression> arguments)
+    {
+      Contract.Ensures(Contract.Result<NewExpression>() != null);
```
SergeyTeplyakov: We can try to pick this precondition: http://referencesource.microsoft.com/#System.Core/Microsoft/Scripting/Ast/NewExpression.cs,230

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.SwitchCase.cs
SergeyTeplyakov: In this case, null or empty `testValues` are invalid (http://referencesource.microsoft.com/#System.Core/Microsoft/Scripting/Ast/SwitchCase.cs,102):

`Contract.Requires(testValues != null);`
`Contract.Requires(testValues.Any());`

Not sure what approach is better: with one or two statements...

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.SwitchExpression.cs
SergeyTeplyakov: Actually, I'm not sure about `Contract.ForAll` because it is costly and actully should be added in many other places to be consistent.

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.TypeBinaryExpression.cs
```diff
@@ -52,5 +52,27 @@ public Type TypeOperand
...
+    //     The System.Linq.Expressions.TypeBinaryExpression.Expression property of the result.
+    //
+    // Returns:
+    //     This expression if no children are changed or an expression with the updated
+    //     children.
+    [Pure]
+    public TypeBinaryExpression Update(Expression expression)
+    {
+      Contract.Requires(expression != null);
+      Contract.Ensures(Contract.Result<TypeBinaryExpression>() != null);
```
SergeyTeplyakov: From this: http://referencesource.microsoft.com/#System.Core/Microsoft/Scripting/Ast/TypeBinaryExpression.cs,a071b1eaa5acb0f6

`Contract.Requires(TypeOperand != null && !TypeOperand.IsByRef);`

Microsoft.Research/Contracts/System.Core.Contracts/System.Runtime.CompilerServices.DebugInfoGenerator.cs
```diff
@@ -0,0 +1,84 @@
...
+    // Parameters:
+    //   method:
+    //     The lambda expression that is generated.
+    //
+    //   ilOffset:
+    //     The offset within MSIL code at which to mark the sequence point.
+    //
+    //   sequencePoint:
+    //     Debug information that corresponds to the sequence point.
+    public abstract void MarkSequencePoint(LambdaExpression method, int ilOffset, DebugInfoExpression sequencePoint);
```
SergeyTeplyakov: It seems that `sequencePoint` should not be null.

sharwell:
Questions from **Expression.cs**:

* :bulb: The requires lines in `DebugInfo(SymbolDocumentInfo, int, int, int, int)` might be improved by using `startLine >= 1` instead of `startLine > 0`. In the editor extensions this drives home the fact that these are 1-based indexes. (All assuming of course that the editor extensions actually use the requirement in string form.)
* :question: In `Default(Type)`, what happens if `type` is null? It won't fail there, but will it fail when you try to compile the IL?
* :bulb: `Expression.Parameter` (two overloads) can each require `type != typeof(void)`.
* :exclamation: The requirement in `Expression.Property(Expression, PropertyInfo, IEnumerable<Expression>)` that `instance != null` is incorrect.
* :exclamation: The requirement in `Expression.Property(Expression, Type, string)` that `expression != null` is incorrect.
* :exclamation: The requirement in `Expression.Property(Expression, PropertyInfo, Expression[])` that `instance != null` is incorrect.
* :exclamation: The requirement in `Expression.Property(Expression, string, Expression[])` that `instance != null` is incorrect.
* :bulb: `Expression.Switch` (6 overloads) can include a requirement that `cases != null` (could also require non-empty and non-null items if we are doing that)
* :bulb: `Expression.SwitchCase` (2 overloads) can include a requirement that `testValues != null` (could also require non-empty and non-null items if we are doing that)
* :bulb: `Expression.TryCatch(Expression, CatchBlock[])` can require `handlers != null` (could also require non-empty and non-null if we are doing that)
* :bulb: `Expression.TryCatchFinally(Expression, Expression, CatchBlock[])` can require `finally != null || handlers != null` (could also require items in `handlers` to be non-null when `handlers` itself is not null)
* :bulb: `Expression.TryFault(Expression, Expression)` can require `fault != null`
* :bulb: `Expression.TryFinally(Expression, Expression)` can require `finally != null`

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.DynamicExpression.cs
sharwell: > It appears that as a general rule they promote nulls to empty collections, unless that collection is expected to be non-empty. If that's the case they require it be non-null.

:arrow_right: I think this is the answer. For construction and transformation, the library treats null collections as empty collections, which means null is valid anywhere that empty is. Since the fastest path through `ToReadOnly` for the case of an empty collection is when null is provided, it's safe to assume client code leverages this, especially in performance-sensitive code.

:thought_balloon: In a world where System.Collections.Immutable preceded System.Linq.Expressions, it's likely that all of these parameters would be `ImmutableArray<T>` and would require non-default instances to express empty. But we don't live in that world. :smile:

sharwell:
Outstanding work overall @fedotovalex. You'll definitely need to update the items marked with :exclamation: since they are incorrect *and* the documentation explicitly covers the case. Let me know if you have any questions.

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.SwitchExpression.cs
sharwell: I'm worried adding too many conditions will result in unprovable contract warnings in any code trying to use the API. I'm not sure what to do about it. I wish we could use `Contract.Assume` in a precondition block to indicate conditions that are true but shouldn't attempt to be proven. This would allow us to show the conditions in the editor extensions (thus helping developers do things right) without negatively impacting the usability of cccheck. I'll let you decide which of the conditions we identified should be added, and which ones should be omitted.

SergeyTeplyakov: Totally agree with your concern. One one hand, I'm sure that static checker would not be able to prove many of them. On the other hand, using Code Contracts Editor Extensions will allow an end user to see them and validate them manually.

I would say that we should use `NotNullOrEmpty` but ommit `Contract.ForAll`...

About `AssumeRequires`: maybe this is a good idea for new feature...

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.GotoExpression.cs
fedotovalex: Added.

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.MemberExpression.cs
fedotovalex: Unfortunately, neither MemberInfo nor PropertyInfo have IsStatic, which makes it difficult to construct a reasonable precondition. 

The `Member is FieldInfo || Member is PropertyInfo` precondition is redundant because it is an object invariant. So I added it as a postcondition for the Member property and also as a precondition to Expression.MakeMemberAccess.

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.LoopExpression.cs
fedotovalex: Added.

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.SwitchExpression.cs
fedotovalex: For this and all similar cases:

I added `Contract.Requires(cases != null);`. For the methods that accept a params array, I also added `Contract.Requires(cases.Length >= 1);`. I did not add `Contract.Requires(cases.Any());` or `Contract.Requires(Contract.ForAll(cases, case => case != null);` because those cannot be proven by the static checker and it will be constant hassle for application developers. Perhaps we can revisit it when the static checker gets smarter.

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.NewArrayExpression.cs
fedotovalex: Again, this always holds true because Type is an array.

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.NewExpression.cs
fedotovalex: I decided not to do this because it is fairly complicated (and the fact I barely understand the logic doesn't help either). In addition, the precondition would involve a call to constructor.GetParameters().Length which the static checker will unlikely be able to prove.

Microsoft.Research/Contracts/System.Core.Contracts/System.Linq.Expressions.TypeBinaryExpression.cs
fedotovalex: Again insured by object invariant.

Microsoft.Research/Contracts/System.Core.Contracts/System.Runtime.CompilerServices.DebugInfoGenerator.cs
fedotovalex: Added this.

fedotovalex:
@sharwell, thanks for the thorough review.

Most of the items from the *Expression.cs* list are done. Here are some comments:

* An attempt to compile an expression with Default(null) results in ArgumentNullException deep in the code when it tries to insert a null type value into a dictionary, so I added a precondition for type != null. Perhaps Microsoft guys can file an internal ticket to have it checked at runtime.

* I did not include a precondition for `type != typeof(void)` in *Parameter* overloads because it cannot be proven by the static checker in the simplest cases.

`Contract.Assert(typeof(string) != typeof(void));  // CodeContracts: assert unproven`

* In `Expression.Property(Expression, string, Expression[])`, `expression != null` is actually required as the expression is the only source of type information (http://referencesource.microsoft.com/#System.Core/Microsoft/Scripting/Ast/IndexExpression.cs,214, RequiresCanRead implies non-null).

Everything else on the list should be addressed.

brettshearer:
Can we use PEX or similar to automatically generate tests for the framework, watch for exceptions and autogenerate contracts based on these?
We are trying to use ReviewBot internally to similar effect (but in our case, generating contracts in our own code).

sharwell:
> In `Expression.Property(Expression, string, Expression[])`, `expression != null` is actually required as the expression is the only source of type information (http://referencesource.microsoft.com/#System.Core/Microsoft/Scripting/Ast/IndexExpression.cs,214, RequiresCanRead implies non-null).

I was referring to [`Expression.Property(Expression, Type, string)`](http://referencesource.microsoft.com/#System.Core/Microsoft/Scripting/Ast/MemberExpression.cs,237).

fedotovalex:
> For the overload you mentioned, the documentation for the parameter is incorrect. It should probably be updated to remove the suggestion that instance could be null.

Okay, I updated the comment in the code and also submitted feedback on the corresponding MSDN page.

Let me know if there are any other issues I need to address before this PR can move forward.

SergeyTeplyakov:
I think everything is looking good and feature is ready for merging.

Any objections? @fedotovalex, @sharwell?

sharwell:
:+1:

SergeyTeplyakov:
Merged. Thanks a lot, @sharwell, @fedotovalex!

PR #173, opened 2015-07-27
Fix for #172.

### LINKED ISSUE #172: CC1099: Contract extraction failed: async/iterator issue in async without await
I am getting the error `CC1099` on a particular method.  I have narrowed the method down to one specific change which will cause the failure.  The below sample compiled as a console application with rewriting enabled seems to reliably fail.

Error:
> EXEC : error CC1099: Contract extraction failed: async/iterator issue: In method EmptyAsyncTest.Program.Foo, assembly EmptyAsyncTest\EmptyAsyncTest\obj\Debug\EmptyAsyncTest.exe

Code:

    using System;
    using System.Threading.Tasks;

    namespace EmptyAsyncTest
    {
        class Program
        {
            static void Main(string[] args)
            {
            }

            async Task Foo()
            {
                if (DateTime.Now.Day == 0 && 1 == 2)
                {
                }
            }
        }
    }

The following variations do not trigger the bug:
 - Removing `async` and returning `Task.FromResult(false)`
 - Removing `&& 1 == 2`
 - Removing `DateTime.Now.Day == 0`
 - Adding `await Task.Delay(1)`

Environment:
VS 2015 CTP, Code Contracts 1.9.10714.2, Windows 8.1, Project is .Net 4.5, Only rewriting is enabled
Also repro'd by compiling fe3bf15c36923ebb833087d8c8781ce2ae60e6ce and running `foxtrot`.  Same output.

SergeyTeplyakov:
This change fixes #172.

### DIFF
--- a/Foxtrot/Foxtrot/Extraction/ExtractorVisitor.cs
+++ b/Foxtrot/Foxtrot/Extraction/ExtractorVisitor.cs
@@ -2360,6 +2360,14 @@ namespace Microsoft.Contracts.Foxtrot
                             statementIndex = i;
                             return currentBlockIndex;
                         }
+                        
+                        // Roslyn-based compiler will generate AssignmentStatement with
+                        // Pop as a source node (any complex if-statement will do this). 
+                        // This type of node will lead to
+                        // InvalidOperationException in the EvaluateExpression method.
+                        // This if statement avoids the failure and fixes #172.
+                        if (assign.Source.NodeType == NodeType.Pop)
+                            continue;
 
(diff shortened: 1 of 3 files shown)


SergeyTeplyakov:
Gentle ping for code review...

sharwell:
:+1: (Assuming the test passes)

SergeyTeplyakov:
Yep, tests (new and old ones) are good.

nparrish:
Do you have an estimate on when a new release with this fix will be published?

jholovacs:
yikes this just killed my build.  Is this released?

sampa999:
Is this going to be released soon?

PR #180, opened 2015-07-30
Contracts for System.Reflection.CustomAttributeExtensions

fedotovalex:
This adds contracts for the [System.Reflection.CustomAttributeExtensions](https://msdn.microsoft.com/en-us/library/system.reflection.customattributeextensions(v=vs.110).aspx) class, which is new in .NET framework 4.5.

### DIFF
--- a/Microsoft.Research/Contracts/MsCorlib/System.Reflection.CustomAttributeExtensions.cs
+++ b/Microsoft.Research/Contracts/MsCorlib/System.Reflection.CustomAttributeExtensions.cs
@@ -1,0 +1,1231 @@
+// CodeContracts
+// 
+// Copyright (c) Microsoft Corporation
+// 
+// All rights reserved. 
+// 
+// MIT License
+// 
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+// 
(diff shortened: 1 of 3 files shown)


Microsoft.Research/Contracts/MsCorlib/System.Reflection.CustomAttributeExtensions.cs
```diff
@@ -0,0 +1,1231 @@
...
+    //   T:System.NotSupportedException:
+    //     element is not a constructor, method, property, event, type, or field.
+    //
+    //   T:System.TypeLoadException:
+    //     A custom attribute type cannot be loaded.
+    [Pure]
+    public static IEnumerable<T> GetCustomAttributes<T>(this MemberInfo element, bool inherit) where T : Attribute
+    {
+      Contract.Requires(element != null);
+      Contract.Ensures(Contract.Result<IEnumerable<Attribute>>() != null);
```
sharwell: :question: Any reason not to make this use `IEnumerable<T>` instead of `IEnumerable<Attribute>`? My guess is the current form was easier for copy/paste. I'm assuming the static checker doesn't have a problem with variance like this?

sharwell:
I had one question in the diff but overall :+1:

Microsoft.Research/Contracts/MsCorlib/System.Reflection.CustomAttributeExtensions.cs
fedotovalex: This is a copy/paste typo, thanks for catching it! Fixed now.

sharwell:
:shipit:

SergeyTeplyakov:
Ьукпувю

PR #181, opened 2015-07-31
Added supprot for default(T) expression used in contracts.

### LINKED ISSUE #29: Malfromed Contract. Found Requires after assignment

hubuk:
I do not know how to run Foxtrot tests so it would be great if someone can take a look at DefaultExpression.cs test and confirm it is correct.
This should fix #29.

### DIFF
--- a/Foxtrot/Foxtrot/Checker/PostExtractorChecker.cs
+++ b/Foxtrot/Foxtrot/Checker/PostExtractorChecker.cs
@@ -2266,8 +2266,8 @@ namespace Microsoft.Contracts.Foxtrot
                         // F: no idea why this is != null
                         Contract.Assume(assignment.Target.Type != null);
                         if (ue.Operand is Local &&
-                            (assignment.Target.Type.IsPrimitive || assignment.Target.Type is Struct ||
-                             assignment.Target.Type.IsStructural))
+                            (assignment.Target.Type.IsPrimitive || assignment.Target.Type.IsStructural || assignment.Target.Type.IsTypeArgument ||
+                             assignment.Target.Type is Struct || assignment.Target.Type is EnumNode))
                         {
                             targetIsLocal = true;
                         }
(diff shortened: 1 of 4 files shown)


SergeyTeplyakov:
I'll test this. But actually running foxtrot tests is pretty simple: use a test explorer and click Run All for Foxtrot tests. Nothing more.

hubuk:
Got it, thnaks. I will make sure that the test is correct.

SergeyTeplyakov:
I assume tests are passing? In this case i'll merge the PR.

hubuk:
I would love to check, but cannot build solution in VS2015. Also Build.bat reports errors on CodeTools. It was prebuilt previously. Lack of possibility to build the solution on VS2015 is by design or is it something with my VS installation?

sharwell:
> Lack of possibility to build the solution on VS2015 is by design or is it something with my VS installation?

Neither. Some of the projects use a customized build process which is no longer supported in 2015. Right now Visual Studio 2013 is the only supported build and test environment, although you will still need .NET 4.6 installed on the build machine.

Issue #167, opened 2015-07-23
Update buildCC.bat to check return codes


sharwell:
Currently buildCC does not check return codes from the various calls to other scripts. This can lead to two problems:

1. If the code tools fail to build, the installer will still be created but it will use an old version of the code tools and #137 will reappear. The build logs must be manually inspected after each build to make sure it was successful.

2. In some cases ,AppVeyor currently reports the build was successful even when it fails.

When one of the build scripts results in an error, the build should be stopped and the error code propagated back to the caller.

PR #183, opened 2015-08-02
Propagating build script error codes to the top level script

fedotovalex:
This is to address #167.

@sharwell, I noticed you added a NuGet package reference for the WiX toolset and updated CodeTools BuildMSM.xml to refer to the correct toolset (which is great). However, NuGet packages from the top-level nuget.config are not restored prior to building CodeTools. As a result, running buildCC on a freshly checked out copy fails.

I added a line to restore NuGet packages right at the top of buildCC.bat. Let me know if there is a better way to handle this.

### DIFF
--- a/Microsoft.Research/ManagedContract.Setup/buildMSI.bat
+++ b/Microsoft.Research/ManagedContract.Setup/buildMSI.bat
@@ -2,11 +2,14 @@
 
 rem Make sure NuGet packages are restored before proceeding with the build
 ..\..\.nuget\NuGet.exe restore ..\..\CodeContracts.sln
+if errorlevel 1 exit /b 1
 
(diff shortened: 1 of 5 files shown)


sharwell:
> I added a line to restore NuGet packages right at the top of buildCC.bat. Let me know if there is a better way to handle this.

I would just add a line [here](https://github.com/Microsoft/CodeContracts/blob/master/Microsoft.VisualStudio.CodeTools/CodeToolsSetup/buildMSM.bat#L4) like the following:

```
rem Also need to restore packages from the top-level solution before building CodeTools
..\..\.nuget\NuGet.exe restore ..\..\CodeContracts.sln
```

fedotovalex:
I changed it as you suggested.

PR #189, opened 2015-08-07
ccrewrite: properly escaping type names when serializing metadata

### LINKED ISSUE #186: ccrewrite produces an incorrect type name in IteratorStateMachineAttribute with some generic types
This is related to #47. Now that parsing of escaped type names in attributes had been fixed, ccrewrite fails to escape it back when producing the rewritten assembly. It seems to have no runtime effect, but it causes problems if this assembly is referenced by another assembly that is also rewritten. ccrewrite attempts to parse the referenced assembly and fails due to an incorrect type name encoded in the attribute.

Pull request with a fix will follow as soon as I figure out how to add a unit test for it.

fedotovalex:
This fixes #186.

@SergeyTeplyakov, will you be adding a unit test for this or do you want me to look into this?

### DIFF
--- a/Foxtrot/Tests/FoxtrotTests10.csproj
+++ b/Foxtrot/Tests/FoxtrotTests10.csproj
@@ -142,6 +142,14 @@
       <Project>{9D404D0C-0921-4ADD-BC22-A14AAE2CA7E2}</Project>
       <Name>CRASanitizer</Name>
     </ProjectReference>
+    <ProjectReference Include="..\..\System.Compiler\System.CompilerCC.csproj">
+      <Project>{d7e16b38-3893-4eef-847f-a3be807e9546}</Project>
+      <Name>System.CompilerCC</Name>
+    </ProjectReference>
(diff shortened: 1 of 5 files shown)


SergeyTeplyakov:
Maybe I missed something, but it seems that current testing infrastructure is not suited for this test scenario. UnitTests project (Regressions/Foxtrot/UnitTests) is not suited because it rewrites assemblies manually but there is no simple way to rewrite chains of assemblies.

So I'm not sure how to automate this process for this fix. It is possible to add two projects and just emulate this specific case, but I don't think this is a good solution.

Any ideas how to automate such kind of stuff?

fedotovalex:
My idea was to let the test run as usual and then try reading the rewritten assembly back the same way ccrewrite does it.

I just committed a proof of concept on this branch, take a look. It's not a 100% reproduction of the original issue, but it is good enough to catch the problem and to confirm it was indeed fixed.

SergeyTeplyakov:
Solution looks good for me!

Issue #165, opened 2015-07-22
Sql exceptions:  String or binary data would be truncated


ReverseBlade:
I am getting  below errors when running static cheker

CodeContracts: PISON.DataRepo: [cache] SqlCacheModel: SaveChanges failed: An error occurred while updating the entries. See the inner exception for details.
CodeContracts: PISON.DataRepo: validation error: The field Kind must be a string or array type with a maximum length of '100'.
CodeContracts: PISON.DataRepo: validation error: The field Kind must be a string or array type with a maximum length of '100'.
CodeContracts: PISON.DataRepo: validation error: The field Kind must be a string or array type with a maximum length of '100'.
CodeContracts: PISON.DataRepo: validation error: The field Kind must be a string or array type with a maximum length of '100'.
CodeContracts: PISON.DataRepo: Innner exception: An error occurred while updating the entries. See the inner exception for details.
CodeContracts: PISON.DataRepo: Innner exception: String or binary data would be truncated.

jterry75:
I see this too. It appears from looking at the code that in SuggestionMap.cs

```cs
this.Property(t => t.Kind)
      .IsRequired()
      .HasMaxLength(100);
```

is not long enough for methods that have a prefixed namespace that is quite long. Can we increase the length of the requirement in the entity map and corresponding cache database to 400 which is the "method name" length requirement?

PR #192, opened 2015-08-07
Fix for: Issue #165: Sql exceptions: String or binary data would be truncated

jterry75:
Increases the size of the SuggestionMap.Kind string data to reflect the size of the Method Name length requirement.

### DIFF
--- a/Microsoft.Research/Clousot.Cache/Models/Mapping/SuggestionMap.cs
+++ b/Microsoft.Research/Clousot.Cache/Models/Mapping/SuggestionMap.cs
@@ -30,7 +30,7 @@ namespace Microsoft.Research.CodeAnalysis.Caching.Models.Mapping
             // Properties
             this.Property(t => t.Kind)
                 .IsRequired()
-                .HasMaxLength(100);
+                .HasMaxLength(MethodMap.NamePropertyMaxLength);
 
             this.Property(t => t.Message)
                 .IsRequired();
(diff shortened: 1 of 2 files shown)


Microsoft.Research/Clousot.Cache/Models/Mapping/SuggestionMap.cs
```diff
@@ -30,7 +30,7 @@ public SuggestionMap()
             // Properties
             this.Property(t => t.Kind)
                 .IsRequired()
-                .HasMaxLength(100);
+                .HasMaxLength(400);
```
SergeyTeplyakov: Just in case, can we extract this into the constant with small comment that will describe why this size should be used?

sharwell: I'd like to see the constant added to (and used in) [`MethodMap`](https://github.com/Microsoft/CodeContracts/blob/c71ef8f1dee388ae3559c2f92969849c977c2d35/Microsoft.Research/Clousot.Cache/Models/Mapping/MethodMap.cs#L37), and then referenced from this location.

jterry75:
Updated the MethodMap to include a Name constant and then referenced that from the SuggestionMap based on feedback.

sharwell:
:+1:

jterry75:
What else needs to be done to merge this in?

sharwell:
@jterry75 We're in the process of dealing with #59 right now. We should see a nice momentum boost for the project as soon as that's dealt with. :smile:

jterry75:
@sharwell No worries! I am very excited for the Roslyn work.

SergeyTeplyakov:
Merged!

Issue #190, opened 2015-08-07
Impove string-contained code snippets


Dennis-Petrov:
`cresn`/`cesn` code snippets should generate `string` type name:

        Contract.Ensures(!string.IsNullOrEmpty(Contract.Result<string>()));
        Contract.Requires<ArgumentException>(!string.IsNullOrEmpty(arg));

instead of `String`:

        Contract.Ensures(!String.IsNullOrEmpty(Contract.Result<string>()));
        Contract.Requires<ArgumentException>(!String.IsNullOrEmpty(arg));

Also, something like `cresnw`/`cesnw` will be suitable:

        Contract.Ensures(!string.IsNullOrWhitespace(Contract.Result<string>()));
        Contract.Requires<ArgumentException>(!string.IsNullOrWhitespace(arg));

I know, that I could fix/add snippets by myself, but out-of-the-box always better :).

sharwell:
I completely agree!

Want to try your hand at a pull request? The snippets are located [here](https://github.com/Microsoft/CodeContracts/tree/master/Microsoft.Research/ManagedContract.Setup/CSharpCodeSnippets).

Dennis-Petrov:
Sure, I didn't know that I can add something here. :)

Dennis-Petrov:
New snippets are included in RC.

PR #193, opened 2015-08-07
Issue #190: Impove string-contained code snippets

jterry75:
Fixes all String references to be string (lowercase).


SergeyTeplyakov:
Thanks! Looks good for me!

sharwell:
:+1:

jterry75:
It seems like this and #196 are actually orthogonal should we take them both?

PR #196, opened 2015-08-08
Added new snippets.

Dennis-Petrov:
Snippets use string.IsNullOrWhiteSpace() instead of
string.IsNullOrEmpty, as proposed in #190.

### DIFF
--- a/Microsoft.Research/ManagedContract.Setup/CSharpCodeSnippets/contractensuresnotisnullorwhitespace.snippet
+++ b/Microsoft.Research/ManagedContract.Setup/CSharpCodeSnippets/contractensuresnotisnullorwhitespace.snippet
@@ -1,0 +1,25 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<CodeSnippets xmlns="http://schemas.microsoft.com/VisualStudio/2005/CodeSnippet">
+  <CodeSnippet Format="1.0.0">
+    <Header>
+      <SnippetTypes>
+        <SnippetType>Expansion</SnippetType>
+      </SnippetTypes>
+      <Title>Contract.Ensures(!string.IsNullOrWhiteSpace(Contract.Result()))</Title>
+      <Shortcut>cesnw</Shortcut>
+      <Description>Emits an 'ensures' clause specifying result not null or white space</Description>
+      <Author>Jonathan de Halleux</Author>
+    </Header>
+    <Snippet>
+      <Imports>
+        <Import>
+          <Namespace>System.Diagnostics.Contracts</Namespace>
+        </Import>
+      </Imports>
+      <Declarations>
+      </Declarations>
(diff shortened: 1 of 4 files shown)


sharwell:
:bulb: If you add new snippets, you should add their shortcuts to **shortcuts.txt** (in the same folder). They will also need to be added to the `CSharpSnippets` in **Microsoft.Research/ManagedContract.Setup/ManagedContracts.wxs**. If you aren't comfortable with the last one, just open up a new issue and we can address it separately if this gets merged. :smile: 

:memo: This pull request does not supersede (or conflict with) #193.

:thought_balloon: I am leaning slightly towards #193 (which changes `String` to `string` but still uses `IsNullOrEmpty`). I find that most cases in code which use `IsNullOrWhiteSpace` incur a performance cost but gain very little in terms of argument validation. I'll wait for feedback from others though.

Microsoft.Research/ManagedContract.Setup/CSharpCodeSnippets/contractensuresnotisnullorwhitespace.snippet
```diff
@@ -0,0 +1,25 @@
...
+<CodeSnippets xmlns="http://schemas.microsoft.com/VisualStudio/2005/CodeSnippet">
+  <CodeSnippet Format="1.0.0">
+    <Header>
+      <SnippetTypes>
+        <SnippetType>Expansion</SnippetType>
+      </SnippetTypes>
+      <Title>Contract.Ensures(!string.IsNullOrWhiteSpace(Contract.Result()))</Title>
+      <Shortcut>cesnw</Shortcut>
+      <Description>Emits an 'ensures' clause specifying result not null or white space</Description>
+      <Author>Jonathan de Halleux</Author>
```
sharwell: :bulb: You could change this to your name instead. :smile:

Microsoft.Research/ManagedContract.Setup/CSharpCodeSnippets/contractensuresnotisnullorwhitespace.snippet
```diff
@@ -0,0 +1,25 @@
...
+    </Header>
+    <Snippet>
+      <Imports>
+        <Import>
+          <Namespace>System.Diagnostics.Contracts</Namespace>
```
sharwell: :bulb: Rather than use this strategy, you can use a [non-editable literal](https://gist.github.com/sharwell/cb8e5bfd2c1435936c27#file-throwifargumentnull-snippet-L24-L29) to automatically reference `Contract` using the shortest available form.

Dennis-Petrov:
Thanks, I'll fix shorcuts.txt/WiX script a little later. Concerning performance, these snippets, of course, are not replacement for `IsNullOrEmpty` family, and should be used only when white space strings are not allowed.

Issue #179, opened 2015-07-30
Wrong path to sn.exe in Microsoft.CodeContracts.targets


rsuter:
I cannot build a project with CC and signing because the path in `CodeContracts\Microsoft.Research\ManagedContract.Setup\MSBuild\v14.0\Microsoft.CodeContracts.targets` seems to be wrong: 

    $(MSBuildProgramFiles32)\Microsoft SDKs\Windows\v8.1A\bin\NETFX 4.6 Tools\sn.exe

If I check this path on my system, the `sn.exe` is either in 

    $(MSBuildProgramFiles32)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.6 Tools\sn.exe

or 

    $(MSBuildProgramFiles32)\Microsoft SDKs\Windows\v8.1A\bin\NETFX 4.5.1 Tools\sn.exe

Do I need to install something or is this a problem in the targets file?

(Running VS 2015 RTM with 10240 Windows SDK)

yaakov-h:
I just encountered this as well - Windows 8.1u1 with VS2015 RTM

I have the following:
* `C:\Program Files (x86)\Microsoft SDKs\Windows\v8.1A\bin\NETFX 4.5.1 Tools\sn.exe`
* `C:\Program Files (x86)\Microsoft SDKs\Windows\v8.1A\bin\NETFX 4.5.1 Tools\x64\sn.exe`
* `C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.6 Tools\sn.exe`
* `C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.6 Tools\x64\sn.exe`

I assume in 421efc0 this should have been changed to `10.0A` and `4.6` instead of `8.1A` and `4.6`?

(Ping @SergeyTeplyakov.)

hubuk:
@rsuter May it be closed as fixed?

Issue #160, opened 2015-07-21
Getting sn.exe error code 3 with Code Contracts enabled


Aospade:
When I have Runtime Checking and Static Checking enabled and my dll is strongly named I get the following error when trying to build in VS 2015 RTM.

The command ""C:\Program Files (x86)\Microsoft SDKs\Windows\v8.1A\bin\NETFX 4.6 Tools\sn.exe" /Rc "obj\Debug\MyDllProject.dll" "MyStrongKey"" exited with code 3.

If I turn off the runtime and static checking the dll works just fine.   Seems similar to this past problem with code contracts

https://social.msdn.microsoft.com/Forums/en-US/4edaa3b3-50a5-4672-82bd-4ca5b48b7c27/release-156081310-now-live?forum=codecontracts

PR #202, opened 2015-08-11
Fix MSBuild 14 path to sn.exe

yaakov-h:
Fix for #179 (and maybe #160?).

### DIFF
--- a/Microsoft.Research/ManagedContract.Setup/MSBuild/v14.0/Microsoft.CodeContracts.targets
+++ b/Microsoft.Research/ManagedContract.Setup/MSBuild/v14.0/Microsoft.CodeContracts.targets
@@ -291,7 +291,7 @@
     <PropertyGroup>
       <!-- SN is in the Framework SDK tools path, but TargetFrameworkSDKToolsDirectory isn't always available  -->
       <CodeContractsSnExe Condition=" '$(TargetFrameworkSDKToolsDirectory)' != '' ">$(TargetFrameworkSDKToolsDirectory)sn.exe</CodeContractsSnExe>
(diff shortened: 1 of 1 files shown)


sharwell:
I believe this works on some systems but not others.

I need to do a bit more investigating.

yaakov-h:
Alright.

SergeyTeplyakov:
Currently Roslyn is using the same targets: https://github.com/dotnet/roslyn/blob/9362f37f29898abd5bd4f0e6be127ab7a947a806/build/Targets/VSL.Settings.targets#L233

And I thought that the second line should be valid only in the rare cases. But I agree with @sharwell this fix would be partial as well: it will work in one configurations but not work in the others.

yaakov-h:
What configurations would have MSBuild 14.0 without Windows SDK 10.0A? Visual Studio 2015 provides both, as best I can tell, and on my Windows 8.1 machine with VS2008, VS2013 and VS2015, only Windows SDK 10.0A has .NET 4.6 Tools.

I'll have to investigate further why `$(TargetFrameworkSDKToolsDirectory)` isn't being used in my case, but if I build a dummy project with `msbuild` and `/v:diag`, the following line appears in the output:

    TargetFrameworkSDKToolsDirectory = C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.6 Tools\

sharwell:
I checked this on a Windows 10 machine and a Windows 7 machine, and found that with Visual Studio 2015 RTM a copy of **sn.exe** was located as specified in this pull request. I haven't verified the location on AppVeyor, but based on the results so far I'm ready to :+1: this pull request.

yaakov-h:
Just following up here - @SergeyTeplyakov, can you specify any configurations where this would not work, or is this OK?

SergeyTeplyakov:
@yaakov-h I'll check on my working machine later today.

yaakov-h:
As opposed to your broken machine? :)

SergeyTeplyakov:
Yep. I've checked my Windows 8.1 machine and sn is located in the place specified by the PR.

Merging....

PR #203, opened 2015-08-11
Fix false contract assumption encountered on System.Decimal.op_Explicit

yaakov-h:
I don't think this has any actual effect on the release builds, but it stopped me getting an exception here when debugging.

`op_Explicit`, `op_Implicit`, and `op_UnaryNegation` all have just one argument, so the assumption that all operations for `System.Decimal` have two or more arguments is incorrect.

### DIFF
--- a/Microsoft.Research/ControlFlow/ControlFlow.cs
+++ b/Microsoft.Research/ControlFlow/ControlFlow.cs
@@ -8320,33 +8320,56 @@ namespace Microsoft.Research.CodeAnalysis
             // special case Decimal methods to operators
             if (mdDecoder.Equal(declaringType, mdDecoder.System_Decimal))
             {
-                Contract.Assume(args.Count >= 2);
                 switch (mdDecoder.Name(method))
                 {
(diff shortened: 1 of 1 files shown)


SergeyTeplyakov:
Two questions:

1. Can we add `Contract.Assume(args.Count >= 1);` or `Contract.Assume(args.Count == 1);` for `case "op_UnaryNegation": `
2. Should we use `Contract.Assume(args.Count == 2);` instead of `Contract.Assume(args.Count >= 2);`?

I'm not sure about second one, maybe it can break something, but first one seems very reasonable for me.

yaakov-h:
1. Yes, I missed that one.
2. I also thought it was off we were assuming `>=` when it appears that `==` is always valid.

SergeyTeplyakov:
Could you please update the PR to address the first comment?
I think we can live with without fixing second one.

yaakov-h:
Sorry about the delay, it's been nighttime here.

Updated.

PR #207, opened 2015-08-13
Fix #206

### LINKED ISSUE #206: Contracts build is broken due to incomplete declaration of System.Runtime.CompilerServices.ExtensionAttribute
PR #180 introduced build error.
System.Runtime.CompilerServices.ExtensionAttribute does not exist in mscorlib prior to .Net 4.5
The build fails with Sanitizer error.

tom-englert:
#206: System.Runtime.CompilerServices.ExtensionAttribute does not exist in mscorlib prior to .Net 4.5


Microsoft.Research/Contracts/MsCorlib/System.Runtime.CompilerServices.ExtensionAttribute.cs
```diff
@@ -16,6 +16,7 @@
 
 namespace System.Runtime.CompilerServices
 {
+#if NETFRAMEWORK_4_5
```
sharwell: :bulb: If you move this up to above the `using System;`, it won't be removed if someone does Remove and Sort Usings without this directive defined. The `#endif` should be moved to the end of the file at the same time.

tom-englert: It is like this everywhere else, I just wanted to comply to the existing pattern.
With this both ExtensionAttribute definitions look the same.
If we want to change this, we should do a bulk refactoring.

sharwell:
:+1:

PR #209, opened 2015-08-13
Qualify references to fields to work around CS0136 in VS2013

### LINKED ISSUE #208: Broken build due to reformatting
#109 (and others) seem to break the build.
During reformatting a lot of ```this.``` statements have been removed, leaving ambiguous symbols behind:
```error CS0136: A local variable named 'exp' cannot be declared in this scope because it would give a different meaning to 'exp', which is already used in a 'parent or current' scope to denote something else```
I'm compiling with VS2013.

sharwell:
Fixes #208

@gafter Do you know why we needed to do this? I can't tell if this was a spec change between 2013 and 2015, or if the old compiler had a bug that was fixed for Roslyn.

### DIFF
--- a/Microsoft.Research/AbstractInterpretation/Abstract Domains/Array/ArraySegmentation.cs
+++ b/Microsoft.Research/AbstractInterpretation/Abstract Domains/Array/ArraySegmentation.cs
@@ -1761,11 +1761,11 @@ namespace Microsoft.Research.AbstractDomains
             }
 
             Contract.Assert(elements.Count >= 1);
-            Contract.Assert(limits.Count == elements.Count + 1);
+            Contract.Assert(this.limits.Count == elements.Count + 1);
 
             // Fast, syntactic search for the lower bound
             var found = false;
-            for (var i = 0; i < limits.Count - 1; i++)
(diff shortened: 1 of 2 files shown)


gafter:
There is something called the "invariant meaning" rule in C#, which basically says that you cannot have a simple name being used to mean one thing in one local scope and a different thing in a nested local scope. Line 1821 declares a local named `limits` but line 1764 uses that name to mean something else in the enclosing scope.

> 7.6.2.1 Invariant meaning in blocks
> For each occurrence of a given identifier as a full simple-name (without a type argument list) in an expression or declarator, within the local variable declaration space (§3.3) immediately enclosing that occurrence, every other occurrence of the same identifier as a full simple-name in an expression or declarator must refer to the same entity. This rule ensures that the meaning of a name is always the same within a given block, switch block, for-, foreach- or using-statement, or anonymous function.

Roslyn relaxes this rule so that it would only fail if there are two conflicting definitions (in other words uses such as that at line 1764 don't count). The updated spec isn't out yet.

/cc @MadsTorgersen please make sure the revised language spec updates the section 'Invariant meaning in blocks'

PR #210, opened 2015-08-14
Fix an apparent copy-paste bug in ValueAtReturnExpression.Equals

panacekcz:
`Equals` should compare objects of the same type `ValueAtReturnExpression`, not `OldExpression`.
I don't have a test case that would expose this, but it seems obvious.


SergeyTeplyakov:
Looks good! I don't think that the test case is required.

SergeyTeplyakov:
Merged! :+1: Thanks a lot!

PR #213, opened 2015-08-18
Update Foxtrot tests to xUnit.net

sharwell:
* Convert the entire test suite from **TestInputs.xml** form to parameterized xUnit.net tests using `[Theory]`. The result is 646 individually-runnable tests which take approximately 1-2 seconds each.
* Update **appveyor.yml** to build and test this project. Note that AppVeyor is not yet enabled for this repository, but it is enabled for my fork and you can see the build result [here](https://ci.appveyor.com/project/sharwell/codecontracts/build/1.9.10818.54/tests).


SergeyTeplyakov:
That stuff looks fantastic! Will review today more carefully!

Foxtrot/Tests/CRASanitizeTest.cs
```diff
@@ -16,55 +16,44 @@
...
+using Xunit;
+using Xunit.Abstractions;
 
 namespace Tests
 {
-  [TestClass]
   public partial class Test
   {
-    private TestContext testContextInstance;
+    private readonly ITestOutputHelper _testOutputHelper;
```
SergeyTeplyakov: The idea of abstracting output away looks very promising. But can we think a bit about name of this abstraction?

Maybe something like `ITestDiagnosticStream`, `IOutputStream`, `ITestConsole`?

Foxtrot/Tests/CRASanitizeTest.cs
```diff
@@ -16,55 +16,44 @@
...
 
+      testOutputHelper.WriteLine("  Contract Path: {0}", Path.GetFullPath(Path.Combine(path, contractAssemblyName)));
+      testOutputHelper.WriteLine("  Original Path: {0}", Path.GetFullPath(originalPath));
 
       var errors = checker.CheckSurface(Path.Combine(path, contractAssemblyName), originalPath);
 
-      Assert.AreEqual(0, errors, "Found {0} errors in contract reference assembly {1}", errors, contractAssemblyName);
-      Console.WriteLine("... done.");
+      testOutputHelper.WriteLine(stringWriter.ToString());
+      Assert.Equal(0, errors);
```
SergeyTeplyakov: I would suggest to keep original error message. Otherwise it will take some time to figure out why test failed.

Foxtrot/Tests/CRASanitizeTest.cs
```diff
@@ -132,81 +122,107 @@ private void CheckSourceFiles(string v9proj, string v10proj)
...
       var v10errors = onlyInV10.Count();
       foreach (var source in onlyInV10)
       {
-        Console.WriteLine("File {0} only in {1}", source, v10proj);
+        _testOutputHelper.WriteLine("File {0} only in {1}", source, v10proj);
       }
 
-      Assert.AreEqual(0, v9errors, "Found {0} errors in project file {1}", v9errors, v9proj);
-      Assert.AreEqual(0, v10errors, "Found {0} errors in project file {1}", v10errors, v10proj);
+      Assert.Equal(0, v9errors);
```
SergeyTeplyakov: The same true here.

Without diagnostic the only way to find this stuff is to look at stack trace. Proper messages in the asserts always helps to find the root cause faster.

Foxtrot/Tests/ExtractorTest.cs
```diff
@@ -12,47 +12,262 @@
...
+            get
             {
-                try
-                {
-                    string sourceFile = (string)TestContext.DataRow["Name"];
-                    string options = (string)TestContext.DataRow["Options"];
-                    TestContext.WriteLine("FAILED: Name={0} Options={1}", sourceFile, options);
-                }
-                catch { }
+                return Enumerable.Range(0, ExtractorTestData.Count()).Select(i => new object[] { i });
```
SergeyTeplyakov: Is it necessary for xUnit to use `object` in this case? I know that with xUnit you can use `IEnumerable<int>` in this case.

And another question: why this test source returns a sequence of indices? Is it impossible to use `IEnumerable<TestCaseData>` where `TestCaseData` is an actual value at `ExtractorTestData`?

Foxtrot/Tests/ExtractorTest.cs
```diff
@@ -12,47 +12,262 @@
...
+        [Theory]
+        [MemberData("ExtractorTest")]
+        [Trait("Category", "Runtime")]
+        [Trait("Category", "CoreTest")]
+        [Trait("Category", "V4.0")]
+        [Trait("Category", "Short")]
+        public void ExtractorFailures(int testIndex)
         {
-            var options = new Options(this.TestContext);
+            Options options = ExtractorTestData.ElementAt(testIndex);
```
SergeyTeplyakov: It seems we have O(n^2) complexity here.

Because `ExtractorTestData` generates a sequence each time, calling following code basically generates N^2 number of items.

Foxtrot/Tests/FoxtrotTests10.csproj
```diff
@@ -155,45 +147,14 @@
...
-      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
-    </Content>
-  </ItemGroup>
-  <Import Project="$(MSBuildBinPath)\Microsoft.CSHARP.Targets" />
+  <Import Project="$(MSBuildBinPath)\Microsoft.CSharp.Targets" />
+  <Target Name="EnsureNuGetPackageBuildImports" BeforeTargets="PrepareForBuild">
+    <PropertyGroup>
+      <ErrorText>This project references NuGet package(s) that are missing on this computer. Use NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is {0}.</ErrorText>
+    </PropertyGroup>
+    <Error Condition="!Exists('..\..\packages\xunit.core.2.1.0-beta4-build3109\build\portable-net45+netcore45+wp8+wpa81\xunit.core.props')" Text="$([System.String]::Format('$(ErrorText)', '..\..\packages\xunit.core.2.1.0-beta4-build3109\build\portable-net45+netcore45+wp8+wpa81\xunit.core.props'))" />
```
SergeyTeplyakov: Can we use stable version, but not beta? Or beta is OK?

Foxtrot/Tests/Properties/AssemblyInfo.cs
```diff
@@ -45,3 +46,5 @@
 // by using the '*' as shown below:
 [assembly: AssemblyVersion("1.0.0.0")]
 [assembly: AssemblyFileVersion("1.0.0.0")]
+
+[assembly: CollectionBehavior(CollectionBehavior.CollectionPerAssembly, DisableTestParallelization = true)]
```
SergeyTeplyakov: Just curious: why test parallelization is disabled? Maybe comment would be helpful if there is a reason for this...

Foxtrot/Tests/TestDriver.cs
```diff
@@ -130,7 +132,7 @@ private static void ExtractContracts(string assemblyFile, Options options)
                     ContractNodes contractNodes = null;
                     Extractor.ExtractContracts(
                         assemblyNode, null, null, null, null, out contractNodes,
-                        e => Assert.Fail(e.ToString()), false);
+                        e => Assert.True(false, e.ToString()), false);
```
SergeyTeplyakov: I'm using NUnit all the time, so I'm not super familiar with xUnit. Does this change means that there is no `Fail` method in xUnit? Because `Assert.Faile` is much more descriptive than `Assert.True(false)` IMO.

Foxtrot/Tests/TestDriver.cs
```diff
@@ -159,22 +161,22 @@ internal static string RewriteAndVerify(string sourceDir, string binary, Options
...
 
-        private static int RunProcess(string cwd, string tool, string arguments, bool mustSucceed, string writeBatchFile = null)
+        private static int RunProcess(ITestOutputHelper testOutputHelper, string cwd, string tool, string arguments, bool mustSucceed, string writeBatchFile = null)
         {
-            var capture = RunProcessAndCapture(cwd, tool, arguments, writeBatchFile);
+            var capture = RunProcessAndCapture(testOutputHelper, cwd, tool, arguments, writeBatchFile);
             if (mustSucceed && capture.ExitCode != 0)
             {
-                Assert.AreEqual(0, capture.ExitCode, "{0} returned an errorcode of {1}.", tool, capture.ExitCode);
+                Assert.Equal(0, capture.ExitCode);
```
SergeyTeplyakov: Again. I would love to see original error message.

Foxtrot/Tests/TestDriver.cs
```diff
@@ -448,11 +457,11 @@ private static void CompareExpectedOutput(Options options, OutputCapture capture
                     actual = TrimFilePath(actual, options.SourceFile);
                     expectedLine = TrimFilePath(expectedLine, options.SourceFile);
                 }
-                Assert.AreEqual(expectedLine, actual);
+                Assert.Equal(expectedLine, actual);
             }
             if (expectedIndex < expected.Length)
             {
-                Assert.AreEqual(expected[expectedIndex], null);
+                Assert.Equal(expected[expectedIndex], null);
```
SergeyTeplyakov: `Assert.Null`? I saw you changed similar code before... So should we try to be consistent here as well? 

This stuff is minor, just IMO.

SergeyTeplyakov:
@sharwell Everything looks awesome. I've put some comments and would be ready to merge today...

Foxtrot/Tests/CRASanitizeTest.cs
sharwell: We could, but it probably wouldn't resolve to [Xunit.Abstractions.ITestOutputHelper](https://github.com/xunit/xunit/blob/master/src/abstractions/xunit.abstractions/TestCases/ITestOutputHelper.cs). :wink:

Foxtrot/Tests/CRASanitizeTest.cs
sharwell: Xunit doesn't support this overload unfortunately.

Foxtrot/Tests/CRASanitizeTest.cs
sharwell: Not supported by Xunit for Assert.Equal. :\

Foxtrot/Tests/ExtractorTest.cs
sharwell: This is the only way it worked with test discovery (something to do with type serialization and `Options` is not a trivial type).

Foxtrot/Tests/ExtractorTest.cs
sharwell: We do (have O(n²) complexity). We could get `ExtractorTestData` to return a new array instance instead of using `yield return`, but for the duration of these tests and the length of data returned by `ExtractorTestData` it just didn't matter in the end.

Foxtrot/Tests/FoxtrotTests10.csproj
sharwell: The beta has a number of performance improvements. We've been using it with good success, and I believe both Roslyn and corefx are using this build as well.

Foxtrot/Tests/Properties/AssemblyInfo.cs
sharwell: The tests manipulate static state. MSTest was never parallelized so this attribute is just included to match that behavior. I could add a comment.

Foxtrot/Tests/TestDriver.cs
sharwell: Your assumption is correct. This is the xunit equivalent.

Foxtrot/Tests/TestDriver.cs
sharwell: I would too. :(

Foxtrot/Tests/CRASanitizeTest.cs
SergeyTeplyakov: Oh!! I didn't realize that this name is part of the xUnit:)

Foxtrot/Tests/CRASanitizeTest.cs
SergeyTeplyakov: Hm... Ok :(

Foxtrot/Tests/ExtractorTest.cs
SergeyTeplyakov: Hm.. (and this one third "Hm.." for this review). As I mentioned, I used NUnit and all of this is not an issue there.

But closed for now:(

Foxtrot/Tests/ExtractorTest.cs
SergeyTeplyakov: It seems that parameterized tests are much less usefule in xUnit comparing to NUnit:(. Quadratic complexity is always a bad thing and we already have hundreds of records in the `ExtractorTestData`...

Yield return is much more expressive way to generate test cases, so I don't want to abandon it. I'm just a bit surprised that THE most popular unit test framework for .NET doesn't cover this case out of the box.

Foxtrot/Tests/FoxtrotTests10.csproj
SergeyTeplyakov: Understood.

Foxtrot/Tests/Properties/AssemblyInfo.cs
SergeyTeplyakov: I think comment would be useful. Because anyone can just change this stuff and break everything...

SergeyTeplyakov:
It seems that all my concerns are related to xUnit itself but not to this particular review. I'm really surprise that xUnit lacks some very useful features that I was using in NUnit forever. But it seems that xUnit is a most popular testing framework and de-facto standard for .NET open source code. Even in this review I see a bunch of case that convince me to stay (at least with my own projects) with NUnit, but I don't want to push my personal preference here.

Foxtrot/Tests/ExtractorTest.cs
sharwell: The problem is the way we are using the parameters. `Options` is a mutable class which provides both behavior and data, but the data structures used to parameterize tests *should* be immutable and only contain the test data. This would be a problem regardless of the test framework in use. This is something we can improve on in the future, but for now I'm just happy to see all the tests able to run on AppVeyor and start getting some reports. :smile:

SergeyTeplyakov:
@sharwell I think the only change that I would like to see is a comment regarding disabling of parallel test execution. And I'll merge the PR right after that. 

Is it sounds good for you?

Foxtrot/Tests/ExtractorTest.cs
sharwell: As I mentioned above, the problem is the `Options` class is exceptionally poorly designed as a storage location for test inputs. The complexity is due exclusively to this design and would be corrected in the future if `Options` were made immutable or if we passed parameter values directly to the test methods without wrapping them in `Options`.

sharwell:
@SergeyTeplyakov I added the comment. :smile:

Foxtrot/Tests/ExtractorTest.cs
SergeyTeplyakov: Understood. Good for now, I think.

SergeyTeplyakov:
Awesome! Merging...

SergeyTeplyakov:
Done!

PR #218, opened 2015-08-20
Updating build environment and tools

ndykman:
Basically, this is just modifications to the build.bat, buildCC.bat and other files to better check for version numbers (and use them) in the build. 

Additionally, there is an optional vsVersion argument on the build.bat and buildCC.bat that will force the use of VS12 or VS14 SDK levels in C++ projects. What this mainly allows for is the build to complete if a person only has VS2015 or VS2013 installed, and not require VS2012 to be installed for the C++ toolchain/SDK. 

Of note in this is that if VS12 or VS14 is used as an option, the redistributable Visual C++ runtimes are included with the installer. This hasn't been tested yet (still building VMs for testing installation). 

The only other major change is a copying of old vsixmanifest files to enable them to be restored on a build failure. 

I'm not sure this will be useful to you or not. Also, I can't yet get the build to work with the 2012 toolset, as there are some project files that reference MSBuild 12.0 tasks and features. I may look into this if it is a priority, but using the 2013 build tool seems to work fine (even when using the older VS11 sdk). 

I'm trying to learn more about the project, as I'm a huge fan of the tool. Thanks for all your work in getting this ready for VS2015 and beyond.

### DIFF
--- a/Microsoft.Research/ManagedContract.Setup/MSBuild/v14.0/Microsoft.CodeContracts.targets
+++ b/Microsoft.Research/ManagedContract.Setup/MSBuild/v14.0/Microsoft.CodeContracts.targets
@@ -291,7 +291,7 @@
     <PropertyGroup>
       <!-- SN is in the Framework SDK tools path, but TargetFrameworkSDKToolsDirectory isn't always available  -->
       <CodeContractsSnExe Condition=" '$(TargetFrameworkSDKToolsDirectory)' != '' ">$(TargetFrameworkSDKToolsDirectory)sn.exe</CodeContractsSnExe>
-      <CodeContractsSnExe Condition=" '$(TargetFrameworkSDKToolsDirectory)' == '' ">$(MSBuildProgramFiles32)\Microsoft SDKs\Windows\v8.1A\bin\NETFX 4.6 Tools\sn.exe</CodeContractsSnExe>
+      <CodeContractsSnExe Condition=" '$(TargetFrameworkSDKToolsDirectory)' == '' ">$(MSBuildProgramFiles32)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.6 Tools\sn.exe</CodeContractsSnExe>
     </PropertyGroup>
     <Exec
        Condition="'$(KeyOriginatorFile)' != ''"   


Microsoft.VisualStudio.CodeTools/CodeTools12.sln
```diff
@@ -0,0 +1,431 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
sharwell: :exclamation: This support needs to be implemented without adding new solution files. Currently there are two solutions in the directory; that is already one too many.

Microsoft.VisualStudio.CodeTools/CodeTools14.sln
```diff
@@ -0,0 +1,431 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
sharwell: :exclamation: This support needs to be implemented without adding new solution files. Currently there are two solutions in the directory; that is already one too many.

sharwell:
I looked through this briefly. I like the idea of supporting multiple environments, but not at the cost of maintaining consistency across multiple solutions, projects, and more.

I think we should go back to the drawing board on this and find a solution that reduces the overall amount of files that need to be maintained (e.g. consolidating cases where we already have multiple solutions and/or projects and creating a single build that *just works*).

ndykman:
Agreed this is mainly a stopgap. I'm all for a discussion on making the build process simpler, but in the meantime, this does address a couple of obvious pain points (while noting where others still lie). This does add more cruft by having more solution/project files, no question. But, I didn't think adding a parameterized solution/project files was worth much the much larger overhead for the very little gain.

In the end, I think simplifying the build will depend a lot on what the goals of 2.0 are. I don't have enough experience with the code base to re-scaffold it into a newer project structure, but that's a great thing to work towards. 
 
Thanks for looking over this.

sharwell:
Right now the stopgap is the project must be built with Visual Studio 2013, and .NET 4.6 needs to be installed on the system. These prerequisites aren't the newest, and they aren't small, but at least they are free. :smile:

ndykman:
Very true. I just had a small issue with building if the V11.0 C++ tools weren't there. The other changes are just making sure the version number gets used in all the right places.

ndykman:
Closed for now. Can recreate a better pull request if needed.

PR #219, opened 2015-08-21
Added contract files for FileDialog and related classes

ndykman:
Sorry if this is overlapping with existing contract work, but below would be a set of files to support contracts for the FileDialog.cs and related classes in System.Windows.Forms namespace.

### DIFF
--- a/Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.CommonDialog.cs
+++ b/Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.CommonDialog.cs
@@ -1,0 +1,63 @@
+﻿// CodeContracts
+// 
+// Copyright (c) Microsoft Corporation
+// 
+// All rights reserved. 
+// 
+// MIT License
+// 
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+// 
+// The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+using System;
+using System.Collections;
+using System.ComponentModel;
+using System.Reflection;
+using System.Runtime.InteropServices;
+using System.Diagnostics.Contracts;
+
+namespace System.Windows.Forms
+{
+    public abstract class CommonDialog // : System.ComponentModel.Component
+    {
+        // <summary>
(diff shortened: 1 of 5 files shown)


Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.FileDialog.cs
```diff
@@ -0,0 +1,275 @@
...
+        /// An array of type <see cref="T:System.String"/>, containing the file names of all selected files in the dialog box.
+        /// </returns>
+        public string[] FileNames
+        {
+            get
+            {
+                Contract.Ensures(Contract.Result<string>() != null);
+                return default(string[]);
+            }
+            set
```
fedotovalex: I don't believe FileDialog.FileNames has a setter.

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.FileDialog.cs
```diff
@@ -0,0 +1,275 @@
...
+        public FileDialogCustomPlacesCollection CustomPlaces
+        {
+            get
+            {
+                Contract.Ensures(Contract.Result<FileDialogCustomPlacesCollection>() != null);
+                return default(FileDialogCustomPlacesCollection);
+            }
+            set
+            {
+
```
fedotovalex: Same here, I don't believe this setter exists.

ndykman:
Agreed. Fix checked in.

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.CommonDialog.cs
```diff
@@ -0,0 +1,63 @@
...
+    public abstract class CommonDialog // : System.ComponentModel.Component
+    {
+        // <summary>
+        // Gets or sets an object that contains data about the control.
+        // </summary>
+        // 
+        // <returns>
+        // The object that contains data about the <see cref="T:System.Windows.Forms.CommonDialog"/>.
+        // </returns>
+        // public object Tag { get; set; }
```
SergeyTeplyakov: Do we really need all those commented properties? Is this a common pattern or we can delete them? Just curious, because this looks a bit weird.

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.FileDialog.cs
```diff
@@ -0,0 +1,268 @@
...
+        /// </returns>
+        public string DefaultExt
+        {
+            get
+            {
+                Contract.Ensures(Contract.Result<string>() != null);
+                return default(string);
+            }
+            set
+            {
```
SergeyTeplyakov: I don't think this postcondition makes sense without appropriate precondition in a set method.
The only case this postcondition would be true, if `set` method would not be allowed to set non-null string...

SergeyTeplyakov: Ok. was wrong with my previous comment. You may ignore it. I've looked at the implementation...

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.FileDialogCustomPlacesCollection.cs
```diff
@@ -0,0 +1,44 @@
...
+using System;
+using System.Collections;
+using System.ComponentModel;
+using System.Reflection;
+using System.Runtime.InteropServices;
+using System.Diagnostics.Contracts;
+
+namespace System.Windows.Forms
+{
+    public class FileDialogCustomPlacesCollection
```
SergeyTeplyakov: Do we need this one? It seems empty...

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.FileDialog.cs
```diff
@@ -0,0 +1,268 @@
...
+        /// </summary>
+        /// 
+        /// <returns>
+        /// An array of type <see cref="T:System.String"/>, containing the file names of all selected files in the dialog box.
+        /// </returns>
+        public string[] FileNames
+        {
+            get
+            {
+                Contract.Ensures(Contract.Result<string>() != null);
```
panacekcz: Should be `Contract.Result<string[]>`?

Issue #214, opened 2015-08-19
Missing contracts in System.Windows.Forms.Control


tom-englert:
e.g. DataBindings property is missing "Ensures result != null"

PR #224, opened 2015-08-24
Changes in System.Windows.Forms.Control

ndykman:
In regards to #214. There were some formatting changes to be made more consistent with other files, so the diff is fairly large. Also, there was a renamed file to fix a spelling error in a file name. 

Again, thanks for your consideration.


Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.Control.cs
```diff
@@ -1432,39 +1449,35 @@ public virtual LayoutEngine LayoutEngine
...
+    public IAsyncResult BeginInvoke(Delegate method, params object[] args)
+    {
+        Contract.Requires(method != null);
+        return default(IAsyncResult);
+    }
```
fedotovalex: I think it's safe to add a postcondition for non-null result for both BeginInvoke variants.

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.ControlBindingsCollection.cs
```diff
@@ -0,0 +1,185 @@
...
+    public Binding Add(string propertyName, object dataSource, string dataMember, bool formattingEnabled, DataSourceUpdateMode updateMode, object nullValue, string formatString, IFormatProvider formatInfo)
+    {
+        Contract.Requires(dataSource != null);
+        return default(Binding);
+    }
```
fedotovalex: This and other similar Add overloads can ensure that the result is non-null.

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.ControlBindingsCollection.cs
```diff
@@ -0,0 +1,185 @@
...
+
+    /// <summary>
+    /// Deletes the <see cref="T:System.Windows.Forms.Binding"/> at the specified index.
+    /// </summary>
+    /// <param name="index">The zero-based index of the item to remove. </param><exception cref="T:System.ArgumentOutOfRangeException">The <paramref name="index"/> value is less than 0, or it is greater than the number of bindings in the collection. </exception><filterpriority>1</filterpriority>
+    public new void RemoveAt(int index)
+    {
+        Contract.Requires(index >= 0);
+    }
```
fedotovalex: This should probably have `Contract.Requires(index < this.Count);`. The documentation says it throws ArgumentOutOfRangeException if index value is less than 0, or it is greater than the number of bindings in the collection, so we should try to reflect that.

>

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.Binding.cs
```diff
@@ -0,0 +1,219 @@
...
+   
+    // <summary>
+    // Gets a value indicating whether the binding is active.
+    // </summary>
+    // 
+    // <returns>
+    // true if the binding is active; otherwise, false.
+    // </returns>
+    // <filterpriority>1</filterpriority>
+    public bool IsBinding { get; }
```
SergeyTeplyakov: Should this property declaration be commented out as well?

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.Control.cs
SergeyTeplyakov: Sounds reasonable for me as well.

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.ControlBindingsCollection.cs
SergeyTeplyakov: +1.

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.Binding.cs
ndykman: Agreed.

ndykman:
Fixes were checked in for review.

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.ControlBindingsCollection.cs
ndykman: +1 as well.

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.Cursor.cs
```diff
@@ -0,0 +1,148 @@
...
+            Contract.Requires(!String.IsNullOrEmpty(fileName));
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="T:System.Windows.Forms.Cursor"/> class from the specified resource with the specified resource type.
+        /// </summary>
+        /// <param name="type">The resource <see cref="T:System.Type"/>. </param><param name="resource">The name of the resource. </param>
+        public Cursor(System.Type type, string resource)
+        {
+            Contract.Requires(type != null || (type == null && resource != null));
```
panacekcz: What is the intended meaning here? This is equivalent to `type!=null || resource!=null`.

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.BaseCollection.cs
```diff
@@ -0,0 +1,106 @@
...
+        /// 
+        /// <returns>
+        /// The total number of elements in the collection.
+        /// </returns>
+        
+        public virtual int Count
+        {
+            get
+            {
+                Contract.Ensures(Contract.Result<int>() > 0);
```
SergeyTeplyakov: This postcondition means that the collection could not be empty... Is it correct? Or postcondition should be `>=0`?

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.BindingsCollection.cs
```diff
@@ -0,0 +1,105 @@
...
+    /// </summary>
+    /// 
+    /// <returns>
+    /// The total number of bindings in the collection.
+    /// </returns>  
+    public override int Count
+    {
+        get
+        {
+            Contract.Ensures(Contract.Result<int>() > 0);
```
SergeyTeplyakov: I can't see a reason why we need to add another postcondition to overriden method.

Both static checker and rewriter will use postcondition from base implementation.

Derived class theoretically can provide stronger postconditions, but I don't think that CC support this. But there is no need to duplicate the code, especially because this postcondition looks suspicions to me (I think it should be `Contract.Result<int>() >= 0` instead).

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.BindingsCollection.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }  
+
+    /// <summary>
+    /// Adds the specified binding to the collection.
+    /// </summary>
+    /// <param name="binding">The <see cref="T:System.Windows.Forms.Binding"/> to add to the collection. </param>
+    protected internal void Add(Binding binding)
+    {
+        Contract.Requires(binding != null);
```
SergeyTeplyakov: Can we add a postcondition that NewCount is greater than the old count?

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.BindingsCollection.cs
```diff
@@ -0,0 +1,105 @@
...
+    // protected virtual void ClearCore();
+    
+    /// <summary>
+    /// Deletes the binding from the collection at the specified index.
+    /// </summary>
+    /// <param name="index">The index of the <see cref="T:System.Windows.Forms.Binding"/> to remove. </param>
+    protected internal void RemoveAt(int index)
+    {
+        Contract.Requires(index >= 0);
+        Contract.Requires(index < this.Count);
```
SergeyTeplyakov: Can we add a postcondition that the new count is less then an old count?

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.ControlBindingsCollection.cs
```diff
@@ -0,0 +1,192 @@
...
+
+    /// <summary>
+    /// Creates a <see cref="T:System.Windows.Forms.Binding"/> using the specified control property name, data source, and data member, and adds it to the collection.
+    /// </summary>
+    /// 
+    /// <returns>
+    /// The newly created <see cref="T:System.Windows.Forms.Binding"/>.
+    /// </returns>
+    /// <param name="propertyName">The name of the control property to bind. </param><param name="dataSource">An <see cref="T:System.Object"/> that represents the data source. </param><param name="dataMember">The property or list to bind to. </param><exception cref="T:System.ArgumentNullException">The <paramref name="binding"/> is null. </exception><exception cref="T:System.Exception">The <paramref name="propertyName"/> is already data-bound. </exception><exception cref="T:System.Exception">The <paramref name="dataMember"/> doesn't specify a valid member of the <paramref name="dataSource"/>. </exception><filterpriority>1</filterpriority>
+    public Binding Add(string propertyName, object dataSource, string dataMember)
```
SergeyTeplyakov: I think there is a postcondition that result is not null.

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.ControlBindingsCollection.cs
```diff
@@ -0,0 +1,192 @@
...
+    // <param name="control">The <see cref="T:System.Windows.Forms.IBindableComponent"/> the binding collection belongs to.</param>
+    // public ControlBindingsCollection(IBindableComponent control)
+       
+    /// <summary>
+    /// Adds the specified <see cref="T:System.Windows.Forms.Binding"/> to the collection.
+    /// </summary>
+    /// <param name="binding">The <see cref="T:System.Windows.Forms.Binding"/> to add. </param><exception cref="T:System.ArgumentNullException">The <paramref name="binding"/> is null. </exception><exception cref="T:System.ArgumentException">The control property is already data-bound. </exception><exception cref="T:System.ArgumentException">The <see cref="T:System.Windows.Forms.Binding"/> does not specify a valid column of the <see cref="P:System.Windows.Forms.Binding.DataSource"/>. </exception><filterpriority>1</filterpriority><PermissionSet><IPermission class="System.Security.Permissions.FileIOPermission, mscorlib, Version=2.0.3600.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" version="1" Unrestricted="true"/></PermissionSet>
+    public void Add(Binding binding)
+    {
+        Contract.Requires(binding != null);
```
SergeyTeplyakov: Can we ensure that new count is greater than old count?

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.ControlBindingsCollection.cs
```diff
@@ -0,0 +1,192 @@
...
+    // </summary>
+    // public new void Clear();
+        
+    /// <summary>
+    /// Deletes the specified <see cref="T:System.Windows.Forms.Binding"/> from the collection.
+    /// </summary>
+    /// <param name="binding">The <see cref="T:System.Windows.Forms.Binding"/> to remove. </param><exception cref="T:System.NullReferenceException">The <paramref name="binding"/> is null. </exception><filterpriority>1</filterpriority>
+    public new void Remove(Binding binding)
+    {
+        Contract.Requires(binding != null);
```
SergeyTeplyakov: Can we ensure that old count is greater than new one?

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.Cursor.cs
ndykman: True. The intent was to capture that if type == null, then the resource can not be null.

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.BaseCollection.cs
ndykman: Bug. Good catch.

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.ControlBindingsCollection.cs
ndykman: I will go through and check for postconditions. Thanks!

Microsoft.Research/Contracts/System.Windows.Forms/System.Windows.Forms.BindingsCollection.cs
ndykman: I could. However, I have a question. This isn't part of the public surface (it is internal). It is better not to have this part of the contract assembly? I'm working on another merge and I'm not including protected internal to reduce the surface a bit. 

Thanks again for your kind consideration.

SergeyTeplyakov: Because this method is protected internal it means that it is actually a part of the public surface. 'protected internal' means 'protected or internal' but not 'protected and internal'...

I'm ok to do this in different PR, I just pointed out that if we already adding this precondition we can add a postcondition as well...

ndykman: Argh. That's right.

ndykman:
Looked at and added postconditions. Will review as noted with my other pull requests.

PR #229, opened 2015-08-26
Update the ClousotTests project to use xUnit.net

sharwell:
:memo: This change does *not* fix the test failures because it's not clear how all of them should be handled. Only one test was updated as an example of how to exclude tests from .NET 3.5.

:memo: This change does not enable the tests to run on AppVeyor because they take at least 12 hours and many of them are currently failing.


sharwell:
:warning: This change broke the build for ClousotCacheTests due to some code sharing. I need to update it before it can be merged.

SergeyTeplyakov:
Understood.

tom-englert:
Just did a quick review, and what's working so far looks promising, a big step forward for the maintainability of the clousot tests. Hopefully that will be finished soon. :+1:

sharwell:
@tom-englert Do the ClousotCacheTests currently pass for you?

sharwell:
Well, I fixed the compilation errors in the cache tests. They aren't converted to xUnit as part of this, and it's possible the changes "broke" them.

tom-englert:
I never had them running correctly, not even after rolling back to the very first version - so there is nothing to break. Best would be to convert them to xUnit, too, then we can try to fix them.

SergeyTeplyakov:
Merged. Sorry for long delay:)

Issue #232, opened 2015-08-29
Outdated email address in output


tom-englert:
Today I got this error message: 
```
CodeContracts: TomsToolbox.Wpf: Internal error: Method TomsToolbox.Wpf.Converters.ConfirmedCommandConverter+CommandProxy.#ctor(TomsToolbox.Wpf.Converters.ConfirmedCommandConverter,System.Windows.Input.ICommand) was found in the cache but deserialization failed.
CodeContracts: TomsToolbox.Wpf: The reason why the deserialization failed is Unable to find type ICommand
CodeContracts: TomsToolbox.Wpf: We remove the incorrect entry from the cache and we replace it with a new one. 
If the error persists, please report the tool output to logozzo@microsoft.com
```
**logozzo@microsoft.com** is probably no longer valid.
We should replace all email addresses with a link to the github issues page.

tom-englert:
@mike-barnett there is some error reporting workflow in the ContractAdornments service provider that sends emails to 
```const string CrashMailRecepients = "ccixfb@microsoft.com";```
Is this still a valid address? Will those mails be processed?

mike-barnett:
Dang, I forgot about that. No, we've deleted that alias. What do you think we should replace that with? Is there a way to have it automatically get added to an issue in GitHub?

sharwell:
@mike-barnett We could always [generate a link like this](https://github.com/Microsoft/CodeContracts/issues/new?title=ContractAdornments+Crash+Report&body=Place+additional+details+here...%0A%0A--%0AException+details:%0A```%0ATODO%0A```).

tom-englert:
@mike-barnett I've already sent a couple of crash reports :grin:

mike-barnett:
@sharwell: can we make it so that it just adds to an existing issue? (Just because there probably will be a lot of them and hopefully they share a common root cause.) @tom-Englert: Unlikely that anyone has seen them...

sharwell:
@mike-barnett At least initially I think I would prefer closing duplicates to reopening the same issue over and over.

sharwell:
@mike-barnett I'm not sure either. Automated issue reports will be a new one for a project I'm on.

tom-englert:
@mike-barnett would it be possible to reactivate ccixfb@microsoft.com as forwarder to us? I get this message at least once a day, but all the information will be lost if the mail gets trashed.

mike-barnett:
[...]

PR #233, opened 2015-08-30
Replace email addresses in error message with the github issues page.

tom-englert:
Fix issue #232

### DIFF
--- a/Microsoft.Research/ClousotMain/Caching/CacheManager.cs
+++ b/Microsoft.Research/ClousotMain/Caching/CacheManager.cs
@@ -1121,7 +1121,8 @@ namespace Microsoft.Research.CodeAnalysis
 
                     this.ReplayOutput.WriteLine("The reason why the deserialization failed is {0}", e.Message);
                     this.ReplayOutput.WriteLine(
-                        "We remove the incorrect entry from the cache and we replace it with a new one. If the error persists, please report the tool output to logozzo@microsoft.com");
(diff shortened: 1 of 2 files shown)


sharwell:
We may have to change these again very soon, but at least for today it's correct. :grin:

Microsoft.Research/ClousotMain/Caching/CacheManager.cs
```diff
@@ -1121,7 +1121,8 @@ private void InstallAssumes(InferredExpr<Field, Method> inferredExpr, Method met
 
                     this.ReplayOutput.WriteLine("The reason why the deserialization failed is {0}", e.Message);
                     this.ReplayOutput.WriteLine(
-                        "We remove the incorrect entry from the cache and we replace it with a new one. If the error persists, please report the tool output to logozzo@microsoft.com");
+                        "We remove the incorrect entry from the cache and we replace it with a new one. "
+                        + "If the error persists, please report the tool output at https://github.com/Microsoft/CodeContracts/issues");
```
SergeyTeplyakov: If we're going to change this in the future, maybe it would be useful to extract this path into some constant and not hard-code it in the message?

sharwell: @SergeyTeplyakov I'm fairly certain we are going to change it *exactly* one more time, but it still could be a good idea to separate it. :+1:

PR #234, opened 2015-08-30
Move System.Compiler.xml so it stops changing during every build

sharwell:
:tada:


SergeyTeplyakov:
Awesome fix!

System.Compiler/System.CompilerCC.csproj
```diff
@@ -228,6 +212,16 @@
...
   </PropertyGroup>
+  <PropertyGroup>
+    <!--
+      Make sure any documentation comments which are included in code get checked for syntax during the build, but do
+      not report warnings for missing comments.
+
+      CS1591: Missing XML comment for publicly visible type or member 'Type_or_Member'
+    -->
+    <DocumentationFile>$(OutputPath)$(AssemblyName).xml</DocumentationFile>
+    <NoWarn>$(NoWarn),1591</NoWarn>
```
SergeyTeplyakov: I see that previous code has another "nowarn" for 1699. Do we need it as well?

sharwell: All instances of that warning were resolved sometime in the past.

SergeyTeplyakov: Understood. Merged.

PR #236, opened 2015-09-01
Contract for TypeConverter.ConvertFrom(object)

Nathan-022:
As far as I can tell this is a valid contract, it doesn't appear that ConvertFrom should ever return null - it either successfully returns the conversion or throws a NotSupportedException

### DIFF
--- a/Microsoft.Research/Contracts/System/System.ComponentModel.TypeConverter.cs
+++ b/Microsoft.Research/Contracts/System/System.ComponentModel.TypeConverter.cs
@@ -83,7 +83,7 @@ namespace System.ComponentModel
     // Returns:
     //     true if this converter can perform the conversion; otherwise, false.
     //public virtual bool CanConvertTo(ITypeDescriptorContext context, Type destinationType);
-    //
+
     // Summary:
     //     Converts the given value to the type of this converter.
     //
@@ -97,8 +97,12 @@ namespace System.ComponentModel
     // Exceptions:
     //   System.NotSupportedException:
     //     The conversion cannot be performed.
-    //public object ConvertFrom(object value);
-    //
+    public object ConvertFrom(object value)
+		{
+			Contract.Ensures(Contract.Result<object>() != null);
+			return default(object);
+		}
+    
     // Summary:
     //     Converts the given object to the type of this converter, using the specified
     //     context and culture information.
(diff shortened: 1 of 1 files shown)


ndykman:
Should doesn't mean can never. The following is an example of getting the method to return null. Pathological, but possible. 

```C#
  public class NullTypeConverter : TypeConverter
  {
    public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
    {
        return null;
    }
  }

  [TypeConverter(typeof(NullTypeConverter))]
  public class MyNullClass
  {
  }

  public static class Program
  {
    public static void Main()
    {
      /// Should print true 
      Console.WriteLine(TypeDescriptor.GetConverter(typeof(MyNullClass)).ConvertFrom("foo") == null);
    }
  }
```

brettshearer:
Is there any return point from this sort of problem to what was probably intended in the framework originally?
I know it would be a breaking change – could we somehow make contracts breaking changes at runtime by a ‘strict’ flag in application manifest?

yaakov-h:
According to the Microsoft Reference Source:

`TypeConverter`:
```
        public object ConvertFrom(object value) {
            return ConvertFrom(null, CultureInfo.CurrentCulture, value);
        }

one two, skip a few...

        public virtual object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value) {
            InstanceDescriptor id = value as InstanceDescriptor;
            if (id != null) {
                return id.Invoke();
            }
            throw GetConvertFromException(value);
        }
```

`InstanceDescriptor`:
```
        public object Invoke() {
            object[] translatedArguments = new object[arguments.Count];
            arguments.CopyTo(translatedArguments, 0);
            
            // Instance descriptors can contain other instance
            // descriptors.  Translate them if necessary.
            //
            for(int i = 0; i < translatedArguments.Length; i++) {
                if (translatedArguments[i] is InstanceDescriptor) {
                    translatedArguments[i] = ((InstanceDescriptor)translatedArguments[i]).Invoke();
                }
            }
            
            if (member is ConstructorInfo) {
                return ((ConstructorInfo)member).Invoke(translatedArguments);
            }
            else if (member is MethodInfo) {
                return ((MethodInfo)member).Invoke(null, translatedArguments);
            }
            else if (member is PropertyInfo) {
                return ((PropertyInfo)member).GetValue(null, translatedArguments);
            }
            else if (member is FieldInfo) {
                return ((FieldInfo)member).GetValue(null);
            }
            else {
                Debug.Fail("Unrecognized reflection type in instance descriptor: " + member.GetType().Name);
            }
            
            return null;
        }
```

Thus in the following code, `convertedValue` will be null:
```
using System.ComponentModel;
using System.ComponentModel.Design.Serialization;

namespace TypeConverterExaggeration
{
	class Program
	{
		static void Main(string[] args)
		{
			var descriptor = new InstanceDescriptor(null, null);
			var converter = new TypeConverter();
			var convertedValue = converter.ConvertFrom(descriptor);
		}
	}
}
```

So even if the framework intended `ConvertFrom` to never return null, the framework's own implementation can do so. I don't see any logical way to apply non-null as a postcondition, even if you want your own code to be stricter than the framework.

PR #256, opened 2015-09-29
Created CodeTools solution along with some project files for VS2015.

hubuk:
Updated CodeTools projects to use NuGets for VSSDK libraries.
Added conditional to make build work with VS2015.
This change is required to evaluate/test changes made using VS2015.


sharwell:
> Added conditional to make build work with VS2015.

The build of CodeTools10 already appears to work in Visual Studio 2015 for me. Can you clarify the reason why changes were needed?

sharwell:
@hubuk I submitted #257 as a possible simpler alternative to this. Can you try it out and let me know if it works for you?

hubuk:
Most of the changes has already been included in my PR.
Why create a new pull request?

sharwell:
@hubuk This pull request is more than 8x the size of #257. Rather than point out all the parts that looked unnecessary, it seemed much easier to just create a new pull request for you to review.

hubuk:
Ok, you wrote that "CodeTools10 already appears to work in Visual Studio 2015". Can you elaborate on this? Maybe this the whole change is not necessary?

sharwell:
The build works without changes on my system, presumably because I have the Visual Studio 2012 build tools installed (that's the `v110`). If it's failing for users who only have Visual Studio 2015, then changing that to `v140` should work for them, but then it would break for users building on older systems. As a balance, I updated the projects to automatically select the build tools matching the version of Visual Studio they are building with.

hubuk:
Closing, in favor of simpler solution in #257

PR #257, opened 2015-09-29
Updated CodeTools10 to support building within Visual Studio 2015

sharwell:
This pull request implements an alternate approach to that taken in #256. Rather than create new solution and project files that will need to be maintained (or simply become out-of-date like so many others in this project), this approach updates the existing project to support both Visual Studio 2013 and Visual Studio 2015.

### DIFF
--- a/Microsoft.VisualStudio.CodeTools/CodeToolsSetup/buildMSM.xml
+++ b/Microsoft.VisualStudio.CodeTools/CodeToolsSetup/buildMSM.xml
@@ -55,11 +55,16 @@
     <Platform>Mixed Platforms</Platform>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(VisualStudioVersion)'==''">
+    <VisualStudioVersion Condition="'$(MSBuildToolsVersion)'=='14.0'">14.0</VisualStudioVersion>
+    <VisualStudioVersion Condition="'$(MSBuildToolsVersion)'!='14.0'">12.0</VisualStudioVersion>
+  </PropertyGroup>
+
   <Target Name="BuildSolutionConfig"
     DependsOnTargets="Validate">
     <MsBuild
       Projects="$(SolutionDir)$(SolutionName)"
-      Properties='Configuration=$(BuildConfig);Platform=$(Platform)'
+      Properties='Configuration=$(BuildConfig);Platform=$(Platform);VisualStudioVersion=$(VisualStudioVersion)'
       Targets="Rebuild"
       />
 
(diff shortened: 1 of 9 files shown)


Microsoft.VisualStudio.CodeTools/CodeToolsSetup/buildMSM.xml
```diff
@@ -55,11 +55,16 @@
     <Platform>Mixed Platforms</Platform>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(VisualStudioVersion)'==''">
+    <VisualStudioVersion Condition="'$(MSBuildToolsVersion)'=='14.0'">14.0</VisualStudioVersion>
+    <VisualStudioVersion Condition="'$(MSBuildToolsVersion)'!='14.0'">12.0</VisualStudioVersion>
```
hubuk: This changes the current behavior which is <PlatformToolset>v110</PlatformToolset>

sharwell: It does. The rest of the build today uses Visual Studio 2013, so I updated this line to match. I can include a third line if you think that would be better:

```xml
<VisualStudioVersion Condition="'$(MSBuildToolsVersion)'=='14.0'">14.0</VisualStudioVersion>
<VisualStudioVersion Condition="'$(MSBuildToolsVersion)'=='12.0'">12.0</VisualStudioVersion>
<VisualStudioVersion Condition="'$(VisualStudioVersion)'==''">11.0</VisualStudioVersion>
```

sharwell:
I noticed in #256 that you changed some assembly references to be project references. Should I do the same thing here? It seemed separate from the issue of 2015 support but we could if you want.

hubuk:
Yes, I was gonna to write about this. For some reason I am getting following warning which later causes errors:
1>C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.Common.CurrentVersion.targets(1819,5): warning MSB3245: Could not resolve this reference. Could not locate the assembly "IPropertyPane, Version=1.0.0.0, Culture=neutral, PublicKeyToken=45d902a1b2562755, processorArchitecture=MSIL". Check to make sure the assembly exists on disk. If this reference is required by your code, you may get compilation errors.

Changing to ProjectReference should fix the problem once and for all.

hubuk:
If we concern only about build on VS2013+ then no additional changes are required. :+1: 
@SergeyTeplyakov is it correct? can this change be merged?

SergeyTeplyakov:
@hubuk I'm ready to merge. Everyting looks good to me...

Issue #270, opened 2015-10-20
No `name` stream error and pdb resolution inconsistency


SergeyTeplyakov:
### First issue. Failure itself

Consider following assembly dependency structure:

```
A.dll --> B.dll --> C.dll
```

`A.dll` references `B.dll` that references `C.dll`, both `B.dll` and `C.dll` has appropriate pdb files.

To build `A.dll` ccrewrite (System.Compiler actually) will build transitive closure of all dependencies including all assemblies and pdbs. It means that System.Comipler will open `B.dll` and `C.dll` with appropriate pdbs.

Current System.Compiler implementation works fine if pdb is absent (in this case debugging experience could be poor) and fail if pdb is broken.

But there is one exception for this rule. Library author can decide to remove all public information from the pdb (by using [/PDBSTRIPPED](https://msdn.microsoft.com/en-us/library/y87kw2fd.aspx) key or by calling [PDBCopy with -p option](https://msdn.microsoft.com/en-us/library/windows/hardware/ff553369(v=vs.85).aspx)). In this case pdb would not have debug information, like `/names` stream.

If `C.dll` in our example does not have debug information in it, all the clients (including transitive clients, like `A.dll` will fail during rewriting process with very weird error message `No ``name' stream`.

### Second issue. Incosistency
Another bug in `AssemblyResolver.cs` makes this issue very hard to investigate. Due to incorrect implementation full name for pdb could be computed incorrectly ([AssemblyResolver.cs](https://github.com/Microsoft/CodeContracts/blob/b2021fb80abd1dbddd0ff5f2bb146ebe2acb5f9d/Foxtrot/Foxtrot/AssemblyResolver.cs#L229)):

```csharp
                        if (String.IsNullOrEmpty(ext))
                        {
                            pdbFullName = Path.Combine(directory, Path.GetFileNameWithoutExtension(assemblyName) + ".pdb");
                        }
```

This call will remove folder name and `pdbFullName` would contain only file name. In this case error become transient and can happen, for instance, only during incremental build, but not during full (clean) build.

Supported versions: pre-VS2013 and current one.

PR #271, opened 2015-10-22
Add special logic for pdbs without /names stream

SergeyTeplyakov:
Fix for #270.

I've tested this stuff manually for generated pdb without /names stream, ran all the tests to check regression.

### DIFF
--- a/System.Compiler/Nodes.cs
+++ b/System.Compiler/Nodes.cs
@@ -5609,6 +5609,19 @@ namespace System.Compiler{
       get{return this.metadataImportErrors;}
       set{this.metadataImportErrors = value;}
     }
+      
+    private ArrayList metadataImportWarnings;
+    /// <summary>
+    /// Holds information about non-critical errors.
+    /// For instance, lack of /names stream in the PDB is not a critical error and should be handled
+    /// gracefully.
+    /// </summary>
+    public ArrayList MetadataImportWarnings {
+      get { return this.metadataImportWarnings; }
+      set { this.metadataImportWarnings = value; }
+    }
+
+
 #endif
     protected AttributeList attributes;
     /// <summary>
--- a/System.Compiler/PDBreader/PdbDebugException.cs
+++ b/System.Compiler/PDBreader/PdbDebugException.cs
@@ -39,7 +39,7 @@ namespace Microsoft.Cci.Pdb
   /// The exception that is thrown when pdb does not contains /names stream.
   /// </summary>
   /// <remarks>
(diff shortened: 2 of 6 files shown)


SergeyTeplyakov:
@tom-englert, @sharwell, @hubuk would appreciate some feedback from you, guys:)

System.Compiler/Nodes.cs
```diff
@@ -5609,6 +5609,19 @@ public Module(TypeNodeProvider provider, TypeNodeListProvider listProvider, Cust
...
       set{this.metadataImportErrors = value;}
     }
+      
+    private ArrayList metadataImportWarnings;
+    /// <summary>
+    /// Holds information about non-critical errors.
+    /// For instance, lack of /names stream in the PDB is not a critical error and should be handled
+    /// gracefully.
+    /// </summary>
+    public ArrayList MetadataImportWarnings {
```
sharwell: :bulb: Seems like this could be the following:

```csharp
public List<Exception> MetadataImportWarnings
{
  get;
  set;
}
```

And the backing field `metadataImportWarnings` removed.

System.Compiler/PDBreader/PdbDebugException.cs
```diff
@@ -27,11 +27,25 @@
...
       : base(String.Format(format, args))
     {
     }
   }
+
+  /// <summary>
+  /// The exception that is thrown when pdb does not contains /names stream.
+  /// </summary>
+  /// <remarks>
+  /// Not all pdb-s contains all possible streams required for debugging. For instnace, 'stripped pdb' (see /PDBSTRIPPED at msdn)
```
sharwell: :bulb: 'instance' is misspelled.

sharwell:
I didn't try it out, but LGTM. I left two comments about small things to consider before merging.

System.Compiler/Nodes.cs
SergeyTeplyakov: Totally agree that for any new code this approach should be used. But I tried to stay with existing "design" that's why, IMHO, for consistency reasons, approach with backing field could be better for now. Once we'll clean up the code, all those ugly .NET 1.0 patterns will disappear:)

System.Compiler/PDBreader/PdbDebugException.cs
SergeyTeplyakov: Thanks!

SergeyTeplyakov:
@sharwell Thanks for the review. Unfortunately, no tests for this fix. The only way I can think about them is to add "broken" pdb, but that definitely not the best approach!

Issue #169, opened 2015-07-24
Process is terminated due to StackOverflowException.


JaCraig:
So it looks like the newest version of ccdocgen is throwing a stackoverflowexception. No stacktrace, just the message in the title. Happens on any project that I've built with any sort of complexity that has the "Emit contracts into XML doc files" checked. Killed many a build today.

Also, I figured that I could spend the night putting in a fix. Did a fork, pulled down, opened up Visual Studio... And I got stuck. I have no idea where that app lives in the code base. I found ccdoc10 but I can't figure out which solution that is housed in. Nor when opening up the project file can I find DataStructures10 or ContractMSInternal. Also it's late and I'm getting tired... If you can point me in the right direction, I'll take a crack at it but otherwise I guess I'll just say glhf (good luck, have fun).

sharwell:
Hi @JaCraig,

Thanks for taking the time to report this, and even to investigate it! We are all living with confusing code for now I think. Like many of the executables in Code Contracts, ccdocgen.exe is created by an ILMerge pass that runs separately from the normal build you see in Visual Studio - hence why you weren't able to find it. The primary ILMerge operations [are defined here](https://github.com/Microsoft/CodeContracts/blob/master/Microsoft.Research/ManagedContract.Setup/buildMSI10.xml#L604-L651), and ccdocgen is created from **ccdoc.exe** plus [these libraries](https://github.com/Microsoft/CodeContracts/blob/master/Microsoft.Research/ManagedContract.Setup/buildMSI10.xml#L486-L504).

**Unfortunately the source code for all of this is not in this repository.** All I see is a [big group of binaries](https://github.com/Microsoft/CodeContracts/tree/master/Microsoft.Research/ImportedCCI2).

ContentsMayVary:
Just to confirm, I am also seeing this issue.

JaCraig:
@sharwell I say this with all love and affection as someone who genuinely likes many of the things that you've worked on over the years (ANTLR, Code Contracts, DartVS, etc): But damn, you crazy.

That is one heck of a build process and debugging must be interesting... Are they binaries for a reason? For instance, if I were to say disassemble those DLLs, set up a project tree, etc. for those things would that be bad? Or are they just binaries because the code has been lost to time and the merge is really the only option without some work?

mike-barnett:
[...]

PR #274, opened 2015-10-24
Update imported CCI2 binaries

tom-englert:
This fixes issue #169 
This is not the optimal solution, but this bug is really blocking so I think we can't wait until there is an official release of CCI2.
See the comments in the commits for how this was done.


SergeyTeplyakov:
Will review this stuff this weekend.

SergeyTeplyakov:
Any chances to add test case for this fix? Just curious, I know that in some cases this would be impossible/very complicated...

tom-englert:
@SergeyTeplyakov adding test cases for this would require to setup the same complex harness as for the clousot tests. I added an issue for this ( #277), since we probably should not wait to fix #169 until we have this tests.

tom-englert:
After Mike has updated the CCI2 sources I have rebuilt everything in a new branch to reduce the complexity of the PR, so this is replaced by  #276

PR #276, opened 2015-10-25
Update imported CCI2 binaries, compile CCDoc

tom-englert:
This fixes issue #169

CCDoc now compiles with solution, binary output is still the ImportedCCI folder so deploy should not be affected.


SergeyTeplyakov:
LGTM. @sharwell: what do you think? I know that we keep updating binaries in the repo, but it seems that this will allows us to fix few very nasty issues...

hubuk:
Looks good.
@SergeyTeplyakov @mike-barnett  Is there any long term plan here? Some major parts of Code Contracts depend on CCI so maybe it could be moved to github under .NET Foundation also?

mike-barnett:
I applied for it to move to the.NET Foundation two months ago. If they don't accept out by the time I get back (December 1), then you can just set up a github site, and I'll close down the codeplex one.

Mike

Sent from my Windows Phone
________________________________

Issue #275, opened 2015-10-24
Synchronous postcondition for async generic method


SergeyTeplyakov:
ccrewrite generates special 'async closure', moves postcondition for task-based method there (both with `async` keyword and without it) and 'chains' a call to `CheckPost` from this class with resulting task.

This behavior is critical, because async postconditions should be checked only when the task completes, but the logic is wrong for async postconditions in generic methods:

```c#
class Foo
{
  public static async Task<T> FooAsync() where T: class
  {
    Contract.Ensures(Contract.Result<T>() != null);
    await Task.Delay(42);
    return default(T);
  }
}
```

In this case synchronous postcondition is emitted:

```c#
// after rewriting
class Foo
{
  public static async Task<T> FooAsync() where T: class
  {
    // some stuff generated by C# compiler
    var task = ...

   RewriterMethods.Ensure(task.Result != null, "Contract.Result<T>() != null");
    return default(T);
  }
}
```

SergeyTeplyakov:
This bug is related to the #235 and I'll push PR with both fixes in a couple of days.

DzheiZee:
Using 1.10.20606.1
Still encountering such defect.
Is fix for this defect included into given version (afaik it is the latest released version)

Issue #235, opened 2015-08-31
BadImageFormatException when using Contract.Ensures in async method, that returns task without any await


Dennis-Petrov:
Hi. This code:

    class Program
    {
        static Task<T> Foo<T>(T source)
            where T : class
        {
            Contract.Ensures(Contract.Result<T>() != null);

            return Task.FromResult(source);
        }

        static void Main(string[] args)
        {
            Foo(new object()).Wait();
        }
    }

leads to `BadImageFormatException`, while this one - not:

        static async Task<T> Foo<T>(T source)
            where T : class
        {
            Contract.Ensures(Contract.Result<T>() != null);

            return await Task.FromResult(source);
        }

The only difference is awaiting the task in the second sample, which is totally unnecessary here.  

It seems to me, that using `Contract.Result<T>()` in `Foo` is logically incorrect here too .We're not *awaiting* a task result, we're just *returning* a task. Though, `BadImageFormatException` shouldn't be thrown, and I need a way to point CC to check task result here.

Is there any workarounds instead of `await`?

yaakov-h:
I believe `Contract.Result<T>()` on a method returning `Task<T>` is only valid on `async` methods, but I'd expect a rewriter error, not a runtime error.

SergeyTeplyakov:
@yaakov-h `Contract.Result<T>()` is valid for both: async and non-async methods.

I've check whether my fix (https://github.com/Microsoft/CodeContracts/pull/159) helps this, but I'm not sure that it will.

BTW, please note, that this code fails on both: VS2015 and VS2013.

SergeyTeplyakov:
BTW, please note, that switching to "async" does not help either: in this case there is no `BadImageFormatException` but resulting code is not valid.

For this code:

```c#
class Program
{
    public static async Task<T> Foo<T>(T source)
    where T : class
    {
        Contract.Ensures(Contract.Result<T>() != null);
        await Task.Yield();
        return source;
        //return Task.FromResult(source);
    }

    static void Main(string[] args)
    {
        Foo("foo").Wait();
    }
}
```

ccrewrite generates following code:

```c#
public static Task<T> Foo<T>(T source) where T : class
{
	Program.<Foo>d__0<T> <Foo>d__;
	<Foo>d__.source = source;
	<Foo>d__.<>t__builder = AsyncTaskMethodBuilder<T>.Create();
	<Foo>d__.<>1__state = -1;
[...]

PR #278, opened 2015-10-25
Async postconditions in generic methods

SergeyTeplyakov:
This PR contains fixes for #235 and #275. Main change in this PR was made in `EmitAsyncClosure` class. 

Special generic type mapping was added to correlate potentially different generic arguments from enclosing generic method to closure generic class.


Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
```diff
@@ -132,21 +248,45 @@ public EmitAsyncClosure(Method from, Rewriter rewriter)
...
+                
+                if (method.DeclaringType.IsGeneric)
+                {
+                    return method.DeclaringType.TemplateParameters;
+                }
+
+                return null;
+            };
+
+            var enclosingTemplateParameters = getGenericTypesFrom(from);
```
hubuk: This delegate seems not to be used anywhere else so the calculation can be performed without it.

SergeyTeplyakov: You're right. I thought to extract method but left with "local" function. Will extract function.

SergeyTeplyakov:
Any other comments on this stuff? The change was relatively major, so I would like to get additional feedback on this...

/cc @sharwell

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
```diff
@@ -59,6 +58,121 @@ namespace Microsoft.Contracts.Foxtrot
...
     {
+        /// <summary>
+        /// Class that maps generic arguments of the enclosed class/method the generic arguments of the closure.
+        /// </summary>
+        /// <remarks>
+        /// The problem.
+        /// Original implementation of the Code Contract didn't support async postconditions in generics.
+        /// Here is why:
+        /// Suppose we have following function (in class <code>Foo</code>:
+        /// <code><![CDATA[
```
hubuk: Just out of curiosity. Why do you used CDATA here? is there any difference in just plain text in <code>?

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
```diff
@@ -59,6 +58,121 @@ namespace Microsoft.Contracts.Foxtrot
...
     internal class EmitAsyncClosure : StandardVisitor
     {
+        /// <summary>
+        /// Class that maps generic arguments of the enclosed class/method the generic arguments of the closure.
+        /// </summary>
+        /// <remarks>
+        /// The problem.
+        /// Original implementation of the Code Contract didn't support async postconditions in generics.
+        /// Here is why:
+        /// Suppose we have following function (in class <code>Foo</code>:
```
hubuk: To inline code into a text a <c>Foo</c> form may be used. <code> will move its content to a new line.

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
AlexanderTaeschner: I assume, that the <T> would be parsed as xml tag otherwise.

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
```diff
@@ -59,6 +58,121 @@ namespace Microsoft.Contracts.Foxtrot
     /// </remarks>
     internal class EmitAsyncClosure : StandardVisitor
     {
+        /// <summary>
+        /// Class that maps generic arguments of the enclosed class/method the generic arguments of the closure.
```
hubuk: Is there a "to" missing?
Class that maps generic arguments of the enclosed class/method **to** the generic arguments of the closure.

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
```diff
@@ -59,6 +58,121 @@ namespace Microsoft.Contracts.Foxtrot
...
+        /// Original implementation of the Code Contract didn't support async postconditions in generics.
+        /// Here is why:
+        /// Suppose we have following function (in class <code>Foo</code>:
+        /// <code><![CDATA[
+        /// public static Task<T> FooAsync() where T: class
+        /// {
+        ///     Contract.Ensures(Contract.Result<T>() != null);
+        /// }
+        /// ]]></code>
+        /// In this case, ccrewrite will generate async closure class called <code>Foo.AsyncContractClosure_0&lt;T%gt;</T></code>
```
hubuk: Should be:
In this case, ccrewrite will generate async closure class called \<**c**\>Foo.AsyncContractClosure_0\&lt;T**&**gt;\</**c**\>

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
hubuk: Ok. Probably because of the need to escape all the xml related parts of the code when not using CDATA.

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
```diff
@@ -59,6 +58,121 @@ namespace Microsoft.Contracts.Foxtrot
...
+            }
+            
+            /// <summary>
+            /// Returns associated generic type of the closure class by enclosing generic type (for instance, by
+            /// generic type of the enclosing generic method that uses current closure).
+            /// </summary>
+            /// <remarks>
+            /// Function returns the same argument if the matching argument does not exists.
+            /// </remarks>
+            public TypeNode GetEnclosingTypeParamterByGenericTypeParameter(TypeNode closureType)
```
hubuk: I am wondering if the intention was to name the method that way: GetEnclosingTypeParamterBy**Closure**TypeParameter

yaakov-h: Param**e**ter.

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
hubuk: What if method is generic, but returned task uses declaring type generic parameter?
Something like this:
```C#
class DeclaringType<T> where T : class {
    public async Task<T> MethodAsync<U>() {
        Contract.Ensures(Contract.Result<T>() != null);
    }
}
```
or scenario with nested type:
```C#
class DeclaringType<T> where T : class {
    private class NestedType<U> where U : class {
        public async Task<T> MethodAsync<V>() {
            Contract.Ensures(Contract.Result<T>() != null);
        }
    }
}
```
Will this also work?

SergeyTeplyakov: Not sure. I'll test this code.

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
SergeyTeplyakov: @hubuk Yep, @AlexanderTaeschner is right. Otherwise we would have to encode all this `<` with `&lt;` which is annoying...

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
SergeyTeplyakov: Thanks, will change!

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
SergeyTeplyakov: Yep! Thanks a lot!

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
SergeyTeplyakov: Thank you!

Foxtrot/Driver/Rewriting/EmitAsyncClosure.cs
SergeyTeplyakov: @hubuk Yep, Will change the name! Thanks for the feedback!

hubuk:
@SergeyTeplyakov I have found a few cases for which BadImageFormatException is thrown.
Some of them are reproducible before fix and some were introduced by the fix. I have prepared 192 combinations of class generic arguments and method arguments. With class and struct constraints. I will present a list of failing scenarios ASAP.

hubuk:
There is also some questionable behavior for closure class signature.
```C#
public class OuterClassGeneric<T0> where T0 : class {
  public class InnerGeneric<T1> {
    public static async Task<T0> AsyncTest1(T0 v0, T1 v1) {
      Contract.Ensures(Contract.Result<T0>() != null);
      return await Task.FromResult(v0);
    }
  }
}
```
rewriter generates the following closure method for the method above:
```C#
.class nested private auto ansi '<AsyncTest1>AsyncContractClosure_0`1'<class T0, T1, T1>
	extends [mscorlib]System.Object
```

Second T1 generic type parameter is probably not needed.

hubuk:
Previously CheckPost method content was:
```C#
TaskStatus status = task.Status;
if (status == TaskStatus.RanToCompletion && __ContractsRuntime.insideContractEvaluation <= 4)
{
  try
  {
    __ContractsRuntime.insideContractEvaluation++;
    __ContractsRuntime.Ensures(task.Result != null, null, "Contract.Result<T0>() != null");
  }
  finally
  {
    __ContractsRuntime.insideContractEvaluation--;
  }
}
return task;
```

now it looks like:
```C#
TaskStatus status = task.Status;
if (status == TaskStatus.RanToCompletion)
{
  __ContractsRuntime.Ensures(task.Result != null, null, "Contract.Result<T0>() != null");
}
return task;
```

Is it intended change?

SergeyTeplyakov:
@hubuk About last comment: the chnage is intented, because async postconditions are all about multi threading where thread static checks like `__ContracatsRuntime.insideContractEvaluation` just makes no sense.

SergeyTeplyakov:
@hubuk Thanks a lot for clear test cases, I'll fix them and will update PR before pushing it to master.

SergeyTeplyakov:
@hubuk FIxed everything that you've mentioned. Including two test cases that you've provde.

Once you'll sign off I'll move forward and will prepare new release of the CodeContracts.

hubuk:
It seems that issue with ``` public static Task<T> DefaultValue<T>(T obj)``` method is not reproducible even without the fix. Second issue is fixed by 2163f15.
LGTM :+1:

SergeyTeplyakov:
@hubuk Great. Moving forward to 1.10 release!

Issue #273, opened 2015-10-24
Bad application of visibility rules on method parameter attributes?


roji:
I'm using [Jetbrains Code Annotations](http://blog.jetbrains.com/dotnet/2015/08/12/how-to-use-jetbrains-annotations-to-improve-resharper-inspections/) in a library which also uses Code Contracts. I'm annotating parameters on some public methods with [NotNull] and [CanBeNull] attributes, but the CC rewriter complains about visibility:

```
Member 'JetBrains.Annotations.CanBeNullAttribute.#ctor' has less visibility than the enclosing method '...'
```

Now, it's true that I'm trying to apply an internal attribute (e.g. [NotNull]) to a public method. But why should this be a problem for CC? Shouldn't it be perfectly fine to have an internal attribute meant purely for internal consumption on a public method's parameters?

SergeyTeplyakov:
@roji interesting question:) Can you show simple workable example? I'm really surprised with this behavior...

roji:
Here's a minimal example:

```c#
public class Program
{
    static void Main(string[] args)
    {
    }

    public static void Foo([MyAttr] string s)
    {
        Contract.Requires(s.StartsWith("yo"));
    }
}

internal class MyAttr : Attribute {}
```

Trying to compile this with runtime contract checking fails with the following error:

```
1>c:\projects\Test2\Program.cs(19,13): error CC1038: Member 'Test2.MyAttr.#ctor' has less visibility than the enclosing method 'Test2.Program.Foo(System.String)'.
```

SergeyTeplyakov:
Thanks a lot! This sounds like a bug to me.

SergeyTeplyakov:
@roji I've created a PR with a fix for this bug. Would appreciate if you would be able to review it. (PR is #279).

SergeyTeplyakov:
New PR is #280.

roji:
Am trying to build CC to test this, but am getting the following when trying to build "Abstract Domains":

```
1>------ Build started: Project: Abstract Domains, Configuration: Debug Any CPU ------
1>  Reading assembly 'CodeAnalysis' from 'C:\projects\CodeContracts\Microsoft.Research\CodeAnalysis\bin\Debug\CodeAnalysis.dll' resulted in errors.
1>  	Illegal characters in path.
1>ccrewrite : error : Rewrite aborted due to metadata errors. Check output window
1>  elapsed time: 2135.5723ms
[...]

PR #280, opened 2015-10-26
Exclude attributes from precondition visibility verification

SergeyTeplyakov:
Fix for #273


Foxtrot/Foxtrot/Checker/VisibilityHelper.cs
```diff
@@ -37,6 +41,22 @@ private void ReInitialize(Method asThisMember)
...
+
+            if (declaringType == systemAttributeType.Value || declaringType.BaseType == systemAttributeType.Value)
+            {
+                return true;
+            }
+            
+
+            return IsAttribute(declaringType.BaseType);
+        }
+
```
hubuk: Current code checks declaringType.BaseType twice on recurrence.
Loop version should be simpler:
```C#
private static bool IsAttribute(TypeNode declaringType)
{
    TypeNode currentType = declaringType;
    while (currentType != null)
	{
	    if (currentType == systemAttributeType.Value)
		{
		    return true;
		}
		
		currentType = declaringType.BaseType;
	}
	
	return false;
}
```

AlexanderTaeschner: In the loop version instead of
```
currentType = declaringType.BaseType
```
this should be
```
currentType = currentType.BaseType
```

hubuk: You are right, thanks. Otherwise it might be less efficient than recursive version...

PR #281, opened 2015-10-27
Changed registry access to use 32 bit view for CodeToolsUpdate project.

hubuk:
This change is required to show Code Contracts property page in Visual Studio after the project platform was changed from x86 to Any CPU.


SergeyTeplyakov:
Looks good to me!

PR #282, opened 2015-10-28
Changed path to Program Files directory to a platform dependent X86 version.

hubuk:
CodeToolsUpdate project was looking for MSBuild in a Program Files directory appropriate for its current architecture. MSBuild is located in X86 version of Program Files.

### DIFF
--- a/Microsoft.VisualStudio.CodeTools/CodeToolsUpdate/TargetInfo.cs
+++ b/Microsoft.VisualStudio.CodeTools/CodeToolsUpdate/TargetInfo.cs
@@ -222,27 +222,27 @@ namespace CodeToolsUpdate
         #region Install for a specific msbuild version
         private string TargetsFile(string msbuildVersion)
         {
-            string msbuildRoot = Environment.GetFolderPath(Environment.SpecialFolder.ProgramFilesX86) + "\\MSBuild\\";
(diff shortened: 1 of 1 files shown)


sharwell:
:thought_balloon: Would be nice to see `Path.Combine` used in this method as well, but that should probably be a separate bug.

:+1:

sharwell:
:question: What does CodeToolsUpdate do?

hubuk:
From my first impression I can only say that it registers VS and MSBuild extensions (Task Manager, property pages, post build tasks) in the system during installation.

SergeyTeplyakov:
@hubuk What do you think, should we change to `Path.Combine` or leaving as is?

SergeyTeplyakov:
@sharwell Are you ok with this change? Or you would like to get an answer on your question first?

hubuk:
I can submit this change in a few minutes.

hubuk:
@SergeyTeplyakov @sharwell Changed to Path.Combine.

fedotovalex:
Hey guys, this fails to compile for me because of a missing closing parenthesis [here](https://github.com/Microsoft/CodeContracts/blob/master/Microsoft.VisualStudio.CodeTools/CodeToolsUpdate/TargetInfo.cs#L245).

hubuk:
Thanks, Submitted a fix. #290

SergeyTeplyakov:
@fedotovalex Thanks! I've merged the fix.

PR #291, opened 2015-11-05
Producing contract reference assemblies for .NET Framework 4.6

fedotovalex:
This pull request adds compiling and installing contract reference assemblies specific for .NET Framework 4.6. This pull request does not add any new system contracts for .NET Framework 4.6, but it would be a prerequisite for adding any such contracts.

##### Why do we need 4.6-specific assemblies?
.NET Framework 4.6 adds new overloads to frequently used methods. For example, there are now

```
static string String.Format(IFormatProvider, string, object)
static string String.Format(IFormatProvider, string, object, object)
static string String.Format(IFormatProvider, string, object, object, object)
```

in addition to 

```
static string String.Format(IFormatProvider, string, object[])
```

This causes innocently looking code like this:

```
string str = String.Format(CultureInfo.InvariantCulture, "Value: {0}", x);
Contract.Assert(str != null);
```

to fail static contract verification when targeting .NET Framework 4.6 because the specific overload is not present in .NET 4.5 CRAs.

##### Caveats
Building new contract reference assemblies requires .NET Framework 4.6 and [.NET Framework 4.6 Targeting Pack](http://www.microsoft.com/en-us/download/details.aspx?id=48136) to be installed on the build machine. 

An alternative would be to stuff another 100MB of reference assemblies into `Microsoft.Research\Imported\ReferenceAssemblies`, but [we seem to feel strongly about not doing so](https://github.com/Microsoft/CodeContracts/blob/master/Microsoft.Research/Imported/ReferenceAssemblies/.NETFramework/v4.6/_readme.txt).

### DIFF
--- a/Microsoft.Research/Contracts/Microsoft.VisualBasic.Compatibility/Microsoft.VisualBasic.Compatibility10.vbproj
+++ b/Microsoft.Research/Contracts/Microsoft.VisualBasic.Compatibility/Microsoft.VisualBasic.Compatibility10.vbproj
@@ -164,8 +164,8 @@
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.VisualBasic.targets" />
   <ItemGroup>
-    <BuildTargetList Include="Microsoft\Framework\v3.5\Target;Microsoft\Framework\.NETFramework\v4.0\Target;Microsoft\Framework\.NETFramework\v4.5\Target" />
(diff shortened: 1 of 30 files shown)


SergeyTeplyakov:
Looks good for me, but I would glad to hear more opinions on this topic.

//cc @sharwell, @tom-englert, @hubuk

tom-englert:
:+1:

SergeyTeplyakov:
LGTM.

@sharwell: would like to hear your opinion.

PR #293, opened 2015-11-07
Bugs bug191

SergeyTeplyakov:


SergeyTeplyakov:
Folks, any comments? 
//cc: @fedotovalex

hubuk:
I can confirm, the fix is working. Good job!
In addition this issue is also observable in Foxtrot\Foxtrot\Extraction\ExtractorVisitor.cs, so many tasts are failing in VS2015.

hubuk:
The only place I would like to change is IsBaseContructorCall. This method returns true also when there is this() constructor call. Maybe rename it to IsChainContructorCall or something similar?

SergeyTeplyakov:
@hubuk Sure, will rename it.

Issue #178, opened 2015-07-29
Async postcondition continuation should run synchronously


SergeyTeplyakov:
Currently, ccrewrite is using following approach for checking postcondition:

```c#
// original async method
var asyncClosure = new AsyncClosure_0();
Task<int> task = t_builder.Task;
task.ContinueWith(asyncClosure.CheckPost);
```

Instead of this, following `ContinueWith` overload should be used: `task.ContinueWith(asyncClosure.CheckPost, TaskContinuationOptions.ExecuteSynchronously)`.

PR #294, opened 2015-11-08
Asynchronous postcondition check now synchronous. Fix for #178

SergeyTeplyakov:


SergeyTeplyakov:
@all folks, I would glad to hear some feedback:)

/cc: @sharwell

hubuk:
@SergeyTeplyakov Looks good :+1: 
But I still don't get the NegativeExpectedCondition and behave variable used in tests.
It seems that when behave is false expected result is dictated by NegativeExpectedKind  and NegativeExpectedCondition. But what this variable is for? Where is it changed?

PR #295, opened 2015-11-08
Remove false contract on PropertyChangedEventArgs

yaakov-h:
MSDN documentation for [PropertyChangedEventArgs Constructor (String)](https://msdn.microsoft.com/en-us/library/system.componentmodel.propertychangedeventargs.propertychangedeventargs(v=vs.110).aspx):
 > An [Empty](https://msdn.microsoft.com/en-us/library/system.string.empty(v=vs.110).aspx) value or **null** for the *propertyName* parameter indicates that all of the properties have changed.


SergeyTeplyakov:
LGTM.

PR #306, opened 2015-11-19
Add contracts for System.OperatingSystem

### LINKED ISSUE #304: No Contracts for System.OperatingSystem
Judging by the Reference source:

* Constructor requires `platform` to be in range.
* Constructor requires non-null `version` or throws `ArgumentNullException`
* `Version` is guaranteed non-null
* `ServicePack` is guaranteed non-null
* `VersionString` is guaranteed not null or empty

yaakov-h:
Based on MSDN and Microsoft Reference Source.

Resolves #304.

I'm not sure if I'm doing `[Pure]` on the getters correctly, and if `VersionString` should have one or not - `VersionString` does change an internal field, but it's only used to cache the result. There's no substantial change to the `OperatingSystem` object.

### DIFF
--- a/Microsoft.Research/Contracts/MsCorlib/System.OperatingSystem.cs
+++ b/Microsoft.Research/Contracts/MsCorlib/System.OperatingSystem.cs
@@ -34,6 +34,7 @@ namespace System
 
     public string ServicePack
     {
+      [Pure]
       get
       {
         Contract.Ensures(Contract.Result<string>() != null);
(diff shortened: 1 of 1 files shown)


Microsoft.Research/Contracts/MsCorlib/System.OperatingSystem.cs
```diff
@@ -23,12 +23,42 @@ public class OperatingSystem: ICloneable
...
+        return default(Version);
+      }
+    }
+
+    public string VersionString
+    {
+      get
+      {
+        Contract.Ensures(!string.IsNullOrWhitespace(Contract.Result<string>()));
+        return default(Version);
```
fedotovalex: I think this meant to be `return default(string)`.

PR #318, opened 2015-12-01
Add contracts for System.Version

yaakov-h:
Cross-checked with Reference Source.

I've verified that this compiles, and work as expected for .NET 4.5.

### DIFF
--- a/Microsoft.Research/Contracts/MsCorlib/System.Version.cs
+++ b/Microsoft.Research/Contracts/MsCorlib/System.Version.cs
@@ -25,37 +25,55 @@ namespace System
         [Pure]
         public int Minor
         {
-            get { return default(int); }
+            get {
+              Contract.Ensures(Contract.Result<int>() >= 0);
+              return default(int);
+            }
         }
 
         [Pure]
         public int Major
         {
-            get { return default(int); }
+            get {
+              Contract.Ensures(Contract.Result<int>() >= 0);
+              return default(int);
+            }
         }
 
         [Pure]
         public int Revision
         {
-            get { return default(int); }
+            get {
+              Contract.Ensures(Contract.Result<int>() >= -1);
+              return default(int);
+            }
         }
 
         [Pure]
         public int Build
         {
-            get { return default(int); }
+            get {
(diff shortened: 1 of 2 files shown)


hubuk:
It should be safe to also add the following contracts for properties.
```C#
//for Major and Minor
Contract.Ensures(Contract.Result<int>() >= 0);
//for Build and Revision
Contract.Ensures(Contract.Result<int>() >= -1);
//for MajorRevision and MinorRevision
Contract.Ensures(Contract.Result<short>() >= -1);
```

yaakov-h:
Thanks @hubuk.

Confirmed still compiling.

Microsoft.Research/Contracts/MsCorlib/System.Version.cs
```diff
@@ -22,24 +22,58 @@ namespace System
...
         {
-            get { return default(int); }
+            get {
+              Contract.Ensures(Contract.Result<int>() >= -1);
+              return default(int);
+            }
+        }
+
+        [Pure]
+        public int MajorRevision
```
hubuk: I haven't spotted this before. [Version.MajorRevision](https://msdn.microsoft.com/en-us/library/system.version.majorrevision%28v=vs.110%29.aspx) is of short type.

Microsoft.Research/Contracts/MsCorlib/System.Version.cs
```diff
@@ -22,24 +22,58 @@ namespace System
...
+        {
+            get {
+              Contract.Ensures(Contract.Result<int>() >= -1);
+              return default(int);
+            }
+        }
+
+        [Pure]
+        public int MinorRevision
```
hubuk: Same for [Version.MinorRevision](https://msdn.microsoft.com/en-us/library/system.version.minorrevision%28v=vs.110%29.aspx).

Microsoft.Research/Contracts/MsCorlib/System.Version.cs
yaakov-h: Good catch, I'll fix them up tomorrow morning.

Issue #307, opened 2015-11-19
Contract.Require(x != null) gets translated in "!(x &lt;= null)" in the documentation rewritten by CC


psmacchia:
We have a lot of *require not null* contracts and I noticed that the contract formatting in the XML doc that used to be `csharp="XYZ != null"` is now `csharp="!(XYZ &lt;= null)"`. If I look in the xml generated I get:

    // XML Generated Now     csharp="!(initialSeq &lt;= null)"
    <requires description="initialSeq must not be null" csharp="!(initialSeq &lt;= null)" vb="!(initialSeq &lt;= Nothing)">!(initialSeq &lt;= null)</requires>

    // XML Generated Before  csharp="initialSeq != null"
    <requires description="initialSeq must not be null" csharp="initialSeq != null" vb="initialSeq &lt;&gt; Nothing">initialSeq != null</requires>

Here is the contract declaration in the C# code:

    // Contract declaration in the C# code
    Contract.Requires(initialSeq != null, "initialSeq must not be null");

Here is the IL emitted from VS2015 (Update 1 RC) after being rewritten with 1.9.1:

    IL_0000: ldarg.0
    IL_0001: ldnull
    IL_0002: cgt.un
    IL_0004: ldstr "initialSeq must not be null"
    IL_0009: ldstr "initialSeq != null"
    IL_000e: call void System.Diagnostics.Contracts.__ContractsRuntime::Requires(bool, string, string)
    IL_0013: nop
    IL_0014: nop
    IL_0015: br IL_001a

Here is the IL when compiled with VS2013 + Code Contract 1.7 if it can help: 

    L_0000: ldarg.0 
    L_0001: ldnull 
    L_0002: ceq 
    L_0004: ldc.i4.0 
    L_0005: ceq 
    L_0007: ldstr "initialSeq must not be null"
    L_000c: ldstr "initialSeq != null"
    L_0011: call void System.Diagnostics.Contracts.__ContractsRuntime::Requires(bool, string, string)
    L_0016: nop 
    L_0017: ldarg.1 
    L_0018: ldnull 
    L_0019: ceq 
    L_001b: ldc.i4.0 
    L_001c: ceq

This is likely cause by Roslyn - Roslyn compiler emits `x != null` as `x > null`, whereas the old <= VS2013 compiler emitted it as `(x == null) == 0` as you can see from the VS2013 IL above.

ccdocgen likely needs to be updated to understand that `x > null` also means `x != null`.

[...]

PR #320, opened 2015-12-08
Fix for issue #307

ndykman:
Fairly straightforward, this adds a overriding method to greater than expressions to look for statements like (x > null) to rewrite them as x != null. Seems to address the issue of documentation (xml documents no longer have !(x <= null) as the source code).

### DIFF
--- a/Microsoft.Research/CCDoc/BooleanExpressionHelper.cs
+++ b/Microsoft.Research/CCDoc/BooleanExpressionHelper.cs
@@ -341,5 +341,23 @@ namespace CCDoc {
       }
       return base.Rewrite(conditional);
     }
+
+      public override IExpression Rewrite(IGreaterThan greaterThan)
+      {
+          // Catch use of the x > null idiom used by Roslyn as generated
+          // code for x != null statements.
+          if (ExpressionHelper.IsNullLiteral(greaterThan.RightOperand))
+          {
+              var c = new NotEquality
+              {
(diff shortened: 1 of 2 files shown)


SergeyTeplyakov:
@ndykman Before we got appveyor support, could you please confirm that all the tests are passing.

Except from that, fix looks good to me.

Microsoft.Research/CCDoc/BooleanExpressionHelper.cs
```diff
@@ -341,5 +341,23 @@ internal class ExpressionSimplifier : CodeRewriter {
...
       return base.Rewrite(conditional);
     }
+
+      public override IExpression Rewrite(IGreaterThan greaterThan)
+      {
+          // Catch use of the x > null idiom used by Roslyn as generated
+          // code for x != null statements.
+          if (ExpressionHelper.IsNullLiteral(greaterThan.RightOperand))
+          {
+              var c = new NotEquality
```
SergeyTeplyakov: Just a suggestion, I would suggest to remove local or use more clear name, like `result`, `nonEquality` or something else.

ndykman:
Can do.

Microsoft.Research/CCDoc/BooleanExpressionHelper.cs
SergeyTeplyakov: Just because it is not that important, I'm merging PR. Thanks a lot for contribution!

PR #324, opened 2016-01-02
 Fix broken regression tests (#148)

tom-englert:
#148:

* Maintainability - Remove experimental projects from the solution and drop related tests:
  * Clousot2*
  * ClousotSlicer*
  * Cloudot*
  * ManyClousot
* Clousot tests:
  * Cleanup the test framework
  * Omit .Net3.5 tests, since many of them fail, and fixing them would be a too expensive task for an outdated framework
  * Ensure all remaining ClousotTests succeed
  * Move ClousotCacheTests to XUnit
  * Activate all Clousot tests


SergeyTeplyakov:
Wov, this is a lots of changes. I'm back from vacation and will start reviewing this stuff shortly.

Microsoft.Research/ClousotMain/RegressionOutput.cs
```diff
@@ -333,7 +318,8 @@ protected MethodRegressionOutcomes(Set<string> outcomes) : base(outcomes)
         {
           var outcome = (ProofOutcome)(byte)GetNamedArgOrDefault<int, Local, Parameter, Method, Field, Property, Event, Type, Attribute, Assembly>(mdDecoder, "Outcome", attr);
 
-          string msg = (string)mdDecoder.NamedArgument("Message", attr);
+          // at some places we have "this.x" instead of "x", both should be treated equal.
```
hubuk: What about situation where "this." is a tail of some other name for example "this.watchthis." ?

tom-englert: Since it's removed on both sides, string comparison will still consider both equal.

All this code is only used by the Clousot regression tests, it does not affect the release.
Writing a test that will falsely succeed because of this would be a real challenge (if possible at all), however we had many tests that randomly fail without this workaround, depending on the environment where they run.

hubuk: Thanks, got it.

Microsoft.Research/RegressionTest/ClousotCacheTests/Tests.cs
```diff
@@ -1,197 +1,253 @@
...
+    using System.Linq;
+
+    using Tests;
+
+    using Xunit;
+    using Xunit.Abstractions;
+
     /// <summary>
-    /// Summary description for RewriterTests
+    /// Cache tests for
```
SergeyTeplyakov: It seems that the comment is not finished. "Cahce tests for *static checker*"?

Microsoft.Research/RegressionTest/ClousotTests/Options.cs
```diff
@@ -4,45 +4,46 @@
...
         private const string RelativeRoot = @"..\..\..\..\..\";
         internal const string TestHarnessDirectory = @"Microsoft.Research\RegressionTest\ClousotTestHarness\bin\debug";
-        private static readonly string RootDirectory;
+        protected static readonly string RootDirectory = Path.GetFullPath(RelativeRoot);
 
-        static Options()
-        {
-            RootDirectory = Path.GetFullPath(RelativeRoot);
-        }
+        private readonly string _outDirectory;
```
SergeyTeplyakov: :question:  Personally, I'm OK with this naming convention, but should we change all the naming in one shot using StyleCopAnalyzers instead of getting inconsistent state with several different naming styles?

This won't block the PR, just would like to clarify and get additional oppinions..

Microsoft.Research/RegressionTest/ClousotTests/Sources/TestFrameworkOOB/UserFeedback.cs
```diff
@@ -394,7 +395,11 @@ public IQueryable<T> OfType<T>()
                 throw new NotImplementedException();
             }
 
-            [ClousotRegressionTest] // CCI2 is not seeing requires
+            [ClousotRegressionTest]
+            
+            // Would this extra check make sense? With it the test will succeed, but what would it prove?
```
SergeyTeplyakov: Have no idea:) Let's left this comment there...

SergeyTeplyakov:
Great job, Tom. LGFM!

Microsoft.Research/RegressionTest/ClousotTests/Options.cs
tom-englert: You're right, I'll do a clean up in the next "iteration". 
Should we clean up project by project, as we work on it? Fixing all projects in one step will probably never happen...

SergeyTeplyakov: I guess we can take combined approach like @sharwell: convert a bunch of project using StyleCopAnalyzers and then try to clean up the code (add more comments, kill a dead code, enhance names etc) as we work on them.

sharwell: @tom-englert I prefer not to do optional things like renaming variables until we get a reasonably reliable and automated set of tests in place which cover that code.

hubuk:
@tom-englert Can we check an AppVeyor build before merging?

Microsoft.Research/RegressionTest/ClousotTests/Options.cs
tom-englert: @sharwell this PR changes only the test framework, and I doubt we will ever have tests that test the test framework :smirk:
After this PR all available tests succeed - so that could be the starting point.

tom-englert:
https://ci.appveyor.com/project/tom-englert/codecontracts
@hubuk: However my builds always terminate with a timeout during the unit tests - how did you manage to bypass this?

hubuk:
@tom-englert From my observations, it is related to the load on appveyor build servers. It timeouts most often in the afternoon for me. It much quicker in the morning.
From time to time there are some problems with the build process itself. My [first build](https://ci.appveyor.com/project/hubuk/codecontracts/build/1.9.10113.125) of your commit have blocked. I just simply re-built and it finished on a second attempt.

Issue #336, opened 2016-01-15
Duplicated contracts when ccrewrite is run with out-of-band contract assembly.


hubuk:
Example of this behavior can be observed in TestResults\Out\Deploy\OutOfBand.dll assembly created when running tests from UnitTests project. The rewrite is taking place in Foxtrot\Tests\UnitTests\RewrittenContractTest.cs on line ```x = RunProcess(deploymentDir, foxtrotPath, "/rewrite /assemblyMode=standard /rw:RewriterMethods.dll,TestRewriterMethods /nobox /throwonfailure /contracts:OutOfBand.Contracts.dll /autoRef:- OutOfBand.dll");```
OOB contracts (/contracts: parameter for ccrewrite.exe) seem not to be used in final product during build.

SergeyTeplyakov:
Can we closed it?

hubuk:
Unfortunately no. This is an issue I found when enabled additional post-rewrite check in foxtrot. It is not critical so I will try to address it some other time.

SergeyTeplyakov:
@hubuk Ok. But this one will not block us from the new release? Right?

hubuk:
@SergeyTeplyakov It should not. The problem is not a regression. The only place I found where Out of band contracts are used (```/contracts:``` switch) is in foxtrot tests. Rewriter run by MSBuild is not using this parameter.

SergeyTeplyakov:
@hubuk Excellent. If you know about any issues/PR that you would really like to see in the upcoming release, please note it here. I'm going to publish new rc-version tomorrow.

hubuk:
@SergeyTeplyakov I was able to provide one more fix for rewriter (issue #168) in PR #344. In addition the build is broken due to lack of some types used in new contracts for Task class. Fix for this is available as #343.

Issue #321, opened 2015-12-10
ccrewrite: Rewritten assembly gets runtime error requesting rewrite


yaakov-h:
I found a very weird corner case in my code this morning.

Using the following code:

```
using System;
using System.Collections.Generic;
using System.Diagnostics.Contracts;

namespace ContractGeneratorMethod
{
	class Program
	{
		public static void Main(string[] args)
		{
			Console.WriteLine("A few of my least favourite things:");

			foreach (var type in Things.LeastFavourite.AFewOfMyLeastFavouriteThings)
			{
				Console.WriteLine("  - {0}", type);
			}
		}
	}

	public static class Things
	{
		public static class LeastFavourite
		{
			// With credits to The Chaser's War on Everything
			public static IEnumerable<string> AFewOfMyLeastFavouriteThings
			{
				get
				{
					Contract.Ensures(Contract.Result<IEnumerable<string>>() != null);

					yield return "Prim Julie Andrews";
					yield return "Cute Singing Kiddies";
				}
			}
		}
	}
}
```

Here's the configurations I've tested:

Configuration | Runtime Contract Checking Enabled | The code works
---|---|---
Debug | No | :white_check_mark: 
Release | No | :white_check_mark: 
Debug | Yes | :white_check_mark: 
Release | Yes | :negative_squared_cross_mark: 

In the non-working configuration (Release with Runtime Contract Checking enabled), the code crashes with the following error:

```
[Window Title]
AAAA.vshost.exe - Postcondition failed.

[Main Instruction]
AAAA.vshost.exe - Postcondition failed.

[Content]
Description: An assembly (probably "AAAA") must be rewritten using the code contracts binary rewriter (CCRewrite) because it is calling Contract.Ensures and the CONTRACTS_FULL symbol is defined.  Remove any explicit definitions of the CONTRACTS_FULL symbol from your project and rebuild.  CCRewrite can be downloaded from http://go.microsoft.com/fwlink/?LinkID=169180. 
After the rewriter is installed, it can be enabled in Visual Studio from the project's Properties page on the Code Contracts pane.  Ensure that "Perform Runtime Contract Checking" is enabled, which will define CONTRACTS_FULL.

[Expanded Information]
Stack Trace:
   at System.Diagnostics.Contracts.Contract.Ensures(Boolean condition)
[...]

PR #337, opened 2016-01-15
Fix for issue #321

hubuk:
Problem described in the issue could be reproduced only when compiled with optimizations enabled (this also requires disabled pdb file) on the Roslyn compiler. Test infrastructure was extended to allow specifying optimization and debug type in the test. Additionally, a new check was introduced in the rewriter that makes sure all contract methods were rewritten for cases where there is no oob contracts in use (issue #336).

[Test results before fix](https://ci.appveyor.com/project/hubuk/codecontracts/build/1.9.10115.130/tests)
[Test results after fix](https://ci.appveyor.com/project/hubuk/codecontracts/build/1.9.10115.131/tests)


Foxtrot/Driver/Rewriting/PostRewriteChecker.cs
```diff
@@ -0,0 +1,120 @@
...
+            if (method.Name.Matches(ContractNodes.RequiresName) ||
+                method.Name.Matches(ContractNodes.EnsuresName) ||
+                method.Name.Matches(ContractNodes.EnsuresOnThrowName) ||
+                method.Name.Matches(ContractNodes.InvariantName) ||
+                method.Name.Matches(ContractNodes.OldName) ||
+                method.Name.Matches(ContractNodes.ResultName) ||
+                method.Name.Matches(ContractNodes.ValueAtReturnName) ||
+                method.Name.Matches(ContractNodes.EndContractBlockName))
+            {
+
```
SergeyTeplyakov: Redundant black line.

Foxtrot/Driver/Rewriting/PostRewriteChecker.cs
```diff
@@ -0,0 +1,120 @@
...
+            {
+                return;
+            }
+
+            if (!method.DeclaringType.Name.Matches(ContractNodes.ContractClassName))
+            {
+                return;
+            }
+
+            if (method.DeclaringType.DeclaringModule.Name != "mscorlib" &&
```
SergeyTeplyakov: Could you please, put a comment, why two assemblies are exceptional?

Foxtrot/Foxtrot/Extraction/ExtractorVisitor.cs
```diff
@@ -1453,6 +1453,33 @@ public TypeNode ClosureClass
...
             }
 
+            public override void VisitReturn(Return Return)
+            {
+                Expression source = Return.Expression;
+                Construct sourceConstruct = source as Construct;
+
+                if (sourceConstruct != null)
+                {
+                    if (sourceConstruct.Type != null && sourceConstruct.Type.Name.Name.StartsWith(this.closureTag))
```
SergeyTeplyakov: Is it possible that `Type.Name` or `Type.Name.Name` will return null?

Foxtrot/Foxtrot/Extraction/ExtractorVisitor.cs
```diff
@@ -1453,6 +1453,33 @@ public TypeNode ClosureClass
                 }
             }
 
+            public override void VisitReturn(Return Return)
```
SergeyTeplyakov: I think better argument name would be `@return` or `returnStatement`, because `Return` violates common .NET naming guidelines...

Foxtrot/Driver/Rewriting/PostRewriteChecker.cs
```diff
@@ -0,0 +1,120 @@
+﻿using System;
+using System.CodeDom.Compiler;
+using System.Collections.Generic;
+using System.Compiler;
+using System.Linq;
+using System.Text;
+
+namespace Microsoft.Contracts.Foxtrot
+{
+    public sealed class PostRewriteChecker : Inspector
```
SergeyTeplyakov: I think, additional comment would be helpful that will explain the whole purpose of the class and what problem it solves...

Foxtrot/Driver/Rewriting/PostRewriteChecker.cs
```diff
@@ -0,0 +1,120 @@
...
+    {
+        private Member currentMember;
+        private readonly Action<CompilerError> handleError;
+
+        public PostRewriteChecker(Action<CompilerError> handleError)
```
SergeyTeplyakov: `Contract.Requirs(handleError != null`?

Foxtrot/Driver/Rewriting/PostRewriteChecker.cs
```diff
@@ -0,0 +1,120 @@
...
+        {
+            this.handleError = handleError;
+        }
+
+        public override void VisitProperty(Property property)
+        {
```
SergeyTeplyakov: Does it means that each `Visit` method can simply return if `this.currentMember` is not `null`?

Foxtrot/Driver/Rewriting/PostRewriteChecker.cs
```diff
@@ -0,0 +1,120 @@
...
+
+            base.VisitMethod(method);
+
+            if (exchanged)
+            {
+                this.currentMember = null;
+            }
+        }
+
+        public override void VisitMemberBinding(MemberBinding memberBinding)
```
SergeyTeplyakov: This code is not that obvious (as any code that has a bunch of visitors:) ), so I would suggest to add a short description of what problem this code is trying to solve in high level.

Foxtrot/Driver/Rewriting/PostRewriteChecker.cs
```diff
@@ -0,0 +1,120 @@
...
+            if (exchanged)
+            {
+                this.currentMember = null;
+            }
+        }
+
+        public override void VisitEvent(Event evnt)
+        {
+            bool exchanged = false;
+            if (this.currentMember == null)
```
SergeyTeplyakov: I'm not 100% sure what `VisitEvent` and `VisitProperty` methods are for... 

I assume that this behavior is clear if you're stepping through the debugger, but reading this code is a bit confusing...

Foxtrot/Driver/Program.cs
```diff
@@ -1333,25 +1333,25 @@ private static int PEVerify(string assemblyFile)
...
-                if (options.contracts == null || options.contracts.Count <= 0) contractFileName = null;
+            if (options.contracts == null || options.contracts.Count <= 0) contractFileName = null;
 
-                if (options.contracts != null &&
-                    !options.contracts.Exists(
-                        name => name.Equals(assemblyNode.Name + ".Contracts.dll", StringComparison.OrdinalIgnoreCase)))
-                    contractFileName = null;
+            if (options.contracts != null &&
+                !options.contracts.Exists(
+                    name => name.Equals(assemblyNode.Name + ".Contracts.dll", StringComparison.OrdinalIgnoreCase)))
```
SergeyTeplyakov: I known this is an old code, but for such a complex `if` statement that we're touching anyway, I would suggest to add curly braces..

Foxtrot/Foxtrot/Extraction/ExtractorVisitor.cs
hubuk: This is a name taken from base method. returnStatement seems to be best choice. Will chnage

Foxtrot/Driver/Rewriting/PostRewriteChecker.cs
hubuk: Contracts may be defined in event and property accessors too. This code checks whether rewriter got it right in these places. Basically we only need VisitMemberBinding method to check correctness of the result. Three other methods (VisitEvent, VisitMethod, VisitProperty) provides a diagnostic information which memeber contains an invalid contract.

SergeyTeplyakov: I meant that comment would be helpful to describe the high level behavior. For instance, first if block is clear what is you doing there, but second if block with `this.currentMember = null` is not clear. I can't get the intend:((

hubuk: Yeah, I will add all the comments to the code. Hopefully everything will be much more clear.

Foxtrot/Foxtrot/Extraction/ExtractorVisitor.cs
hubuk: Probably not in normal situation. If either of them is null it is a signal that the problem is somewhere else.

Foxtrot/Driver/Rewriting/PostRewriteChecker.cs
```diff
@@ -0,0 +1,169 @@
...
+        ///     Initializes a new instance of the <see cref="PostRewriteChecker"/> class with
+        ///     a reference to an action responsible for handling reported errors.
+        /// </summary>
+        /// <param name="handleError">
+        ///     An action that will be called for each error occurred.
+        /// </param>
+        public PostRewriteChecker(Action<CompilerError> handleError)
+        {
+            Contract.Requires(handleError != null);
+            Contract.Ensures(object.ReferenceEquals(this.handleError, handleError));
```
SergeyTeplyakov: I'm fine with this postcondition, but I think it's slightly too much:))))

Foxtrot/Driver/Rewriting/PostRewriteChecker.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// </param>
+        /// <remarks>
+        ///     This method reports error for each call to Code Contracts' method which normally should be rewritten.
+        /// </remarks>
+        public override void VisitMemberBinding(MemberBinding binding)
+        {
+            base.VisitMemberBinding(binding);
+
+            if (object.ReferenceEquals(binding, null) ||
+                this.visitedMembers.Count == 0 ||
```
SergeyTeplyakov: As far as I understand, last object in the `visitedmember` would be method/property or event being analyzed right now? Am I right?

SergeyTeplyakov: So it means that this check ensures that only for those type we need to analyze current bound member represented by `binding`?

SergeyTeplyakov:
LGTM

Foxtrot/Driver/Rewriting/PostRewriteChecker.cs
hubuk: I would expect that this kind of relationship can be inferred by static checker. However in some cases I was forced to add Ensures like this (using equality operator not ReferenceEquals) to mute cccheck warnings. This is more like a habit, which may be based on some early versions of CodeContracts.

Foxtrot/Driver/Rewriting/PostRewriteChecker.cs
hubuk: Yes, last object in ```visitedMembes``` is a member in which the binding we are visiting right now is contained. For example: when method/property/event has an attribute then this attribute's type with all its members may be visited before method/property/event's body. So we need to remember all the members visited using stack.

hubuk:
Squashed multiple commits into one.

Issue #340, opened 2016-01-21
ccrewrite fails to read assembly that has non-escaped comma in nested type names


SergeyTeplyakov:
Related to the #186.

If old version of ccrewrite could generate type names that with non-espcaped comman inside when working with Roslyn-generated code.

It means, that instead of `SomeGeneric+Nested<Int32\,String>` it can generate `SomeGeneric+Nested<Int32,String>`. But actually both cases are correct and metadata reader should deal with both of them smoothly.

tom-englert:
@SergeyTeplyakov with this fix we get crashes in ccrewrite for generic types:

the class with it's attribute looks something like this:
```
[Export(typeof(IEntityToDtoMapping<TestOrder, Modules.TestOrder.ServerAccess.TestOrder, IPersistenceContext>))]
private class Mapping : EntityToDtoMapping<TestOrder, Modules.TestOrder.ServerAccess.TestOrder, IPersistenceContext>
{
```
which will generate a type name like this:
```

SergeyTeplyakov:
Could you provide a small repo solution that I can try? I can revert the change if solution would be complicated...

tom-englert:
I have a fix already in #445, just looking how to add a test for this scenario. Did you have a test for the original issue?

SergeyTeplyakov:
Yep, I've faced an issue but that was not a critical and more artificial case. I'll role this change back.

SergeyTeplyakov:
It seems that you fix it in your fork. If you'll create PR I'll merge it.

tom-englert:
#445 is a pull request :smirk:

SergeyTeplyakov:
Oops! Thought it is an old one. Will merge once appveyor will be finished.

tom-englert:
After adding a test I could not find the scenario described here - instead with the code introduced in #341 it was not able to read it at all, so I have reverted the complete fix in PR #446

PR #341, opened 2016-01-21
Fix for #340.

SergeyTeplyakov:
ccrewrite will not fail if metadata has non-escaped comma in the type name.

As I mentioned in bug #340 this bug is tightly related to #186.

Per internal documentation, comma (`,`) can be used in two different cases in the type names: to separate assembly name or to separate generic type arguments. Here is a comment from MetadataReader codebase:

```c#
// A comma may separate a type name from its assembly name or a type argument from
// another type argument.
// If processing non-type argument or a type argument with assembly name,
// process the characters after the comma as an assembly name.
```

So this fix allows ccrewrite to read the metadata that has unescaped commas in type names.

### DIFF
--- a/System.Compiler/Reader.cs
+++ b/System.Compiler/Reader.cs
@@ -4257,6 +4257,18 @@ namespace System.Compiler.Metadata{
             this.currentItr++;
             switch (c) {
               case ',':
+                    // A comma may separate a type name from its assembly name or a type argument from
+                    // another type argument.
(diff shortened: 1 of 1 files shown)


hubuk:
Do we know for sure that an assembly name is always preceded by white space after comma?

SergeyTeplyakov:
@hubuk Yep. I've copied that comment from Reflection.MetaData sources.

hubuk:
So it looks good. :+1:

PR #335, opened 2016-01-14
Enable automated execution of all unit tests in AppVeyor

asvishnyakov:
You need additional configure AppVeyor account with at least Basic plan (Free plan has a limits, what prevents test successful execution) to enable automated execution

PR #342, opened 2016-01-23
Foxtrot & Clousot tests are disabled because of time limit in AppVeyor free plan

asvishnyakov:
Must be restored when Microsoft upgarde CodeContracts' AppVeyor plan to Basic or high


SergeyTeplyakov:
As you suggested in #335, we can enable at least UnitTests for now.

Issue #168, opened 2015-07-23
Using capturing precondition in async method in struct lead to invalid IL


SergeyTeplyakov:
This bug is related to the #157, but not similar.

Consider following code:

```c#
struct Foo
{
    public int Limit { get; set; }

    public Foo(int limit) : this()
    {
        Limit = 42;
    }

    public async Task<int> FooAsync()
    {
        Contract.Requires(Limit == 42);

        await Task.Delay(42);

        return 42;
    }
}
```

Running this code will lead to `InvalidProgramException` when preconditions are enabled.

CC version: 1.8+. This is not a regression, this stuff never worked properly.

Please note, the if the `Contract.Requires` will use arguments everything would be fine. Only when precondition is using `this` the app will fail.

hubuk:
@SergeyTeplyakov Are you working on this?

SergeyTeplyakov:
@hubuk At least not right now. I hope to merge pull requrests this week, so you can grab it.

hubuk:
Grabbed

PR #344, opened 2016-01-26
Fix for issue #168

hubuk:
Added support for ```AddressOf``` for ```this``` operand.

### DIFF
--- a/System.Compiler/Writer.cs
+++ b/System.Compiler/Writer.cs
@@ -2780,6 +2780,9 @@ namespace System.Compiler{
           }
           this.IncrementStackHeight();
           return;
+        case NodeType.This:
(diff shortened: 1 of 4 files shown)


System.Compiler/Writer.cs
```diff
@@ -2780,6 +2780,9 @@ class MethodSemanticsRowComparer : IComparer{
           }
           this.IncrementStackHeight();
           return;
+        case NodeType.This:
+          this.methodBodyHeap.Write((byte)0x02);
```
SergeyTeplyakov: This fix is so small! But it affects behavior in such a drastic way:)

Looks good to me! Great job!

hubuk: Yes, 121 additions for test and 3 for fix :)

PR #354, opened 2016-01-28
Provided long term solution for broken build.

hubuk:
Added support for .Net reference assemblies located outside repository.
Changed reference assemblies for .Net 4.6 to MSIL in order to get rid of compilation warnings.


SergeyTeplyakov:
Wow! This is awesome!!

PR #358, opened 2016-02-03
Output declarative assemblies to output directory

### LINKED ISSUE #348: Output Declarative assembly to OutDir, instead of IntermediateOutputPath
In the environment I run Code Contracts in, we run the static checker separately to Visual Studio / MSBuild in a parallel distributed environment.

As such, in order to distribute the static checking to different computers, we need to distribute the declarative assemblies from `CodeContractsDeclDir ` (currently defined as `$(IntermediateOutputPath)Decl\`), as well as, strangely the source code in it's original disk location.

It seems logical to me that, whilst the declarative assemblies are technically intermediate, they are also built output since they are also required input for `cccheck`. This is much easier if these are output to the project output directory, rather than somewhere within the obj folder.

I propose moving `CodeContractsDeclDir` to `$(OutDir)\Decl\`, as it becomes easier to distribute the rewritten binaries and the declarative ones, rather than having to go digging around the Obj folder to find the declarative ones. All binaries (Contract Declarative Assembly, Contract Reference Assembly, and Executable Assembly) are then in the same place - somewhere within `$(OutDir)`.

yaakov-h:
Old behaviour: Declarative assemblies live in obj\ somewhere
New behaviour: Declarative assemblies live in bin\ somewhere
(subject to OutputPath being customized by the user)

This is the simplest way to fix #348. Another options would be making this configurable via the properties pane, but I'm not sure how to go about that.

### DIFF
--- a/Microsoft.Research/ManagedContract.Setup/MSBuild/v12.0/Microsoft.CodeContracts.targets
+++ b/Microsoft.Research/ManagedContract.Setup/MSBuild/v12.0/Microsoft.CodeContracts.targets
@@ -9,7 +9,7 @@
   ======================================================================-->
 
   <PropertyGroup>
-    <CodeContractsDeclDir>$(IntermediateOutputPath)Decl\</CodeContractsDeclDir>
+    <CodeContractsDeclDir>$(OutDir)Decl\</CodeContractsDeclDir>
     <CodeContractsContractSubDir>CodeContracts\</CodeContractsContractSubDir>
     <CodeContractsCCRefgenCommand>$(CodeContractsInstallDir)Bin\ccrefgen.exe</CodeContractsCCRefgenCommand>
   </PropertyGroup>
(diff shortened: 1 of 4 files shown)


SergeyTeplyakov:
LGTM. Would like to hear thoughts by other people.

/cc: @hubuk, @tom-englert

hubuk:
Since new directory will be some kind of public now I would like to change name Decl to DeclarativeAssembly or Declarative.
@yaakov-h @SergeyTeplyakov Would it be right or Decl is just fine?

SergeyTeplyakov:
@hubuk My vote would be for better name:)

yaakov-h:
@hubuk Maybe CodeContractsDeclarative?

SergeyTeplyakov:
@yaakov-h :+1:

hubuk:
@yaakov-h CodeContractsDeclarative looks good :+1:

yaakov-h:
Done. Looks like AppVeyor timed out, but it works on my machine :smile:

asvishnyakov:
Does AppVeyor have any options to re-run checks?

hubuk:
@SergeyTeplyakov is probably able to rerun the check.

SergeyTeplyakov:
Restarted the build.

asvishnyakov:
@SergeyTeplyakov Don't forget restart build in other PRs before merge it

yaakov-h:
That looks better :)

PR #359, opened 2016-02-03
Use .NET 4.6 reference assemblies for 4.6.1

yaakov-h:
Also use .NET 4.6 reference assemblies when targetting 4.6/4.6.1 with Visual Studio 2013


SergeyTeplyakov:
LGTM

Daniel-Svensson:
Looks good, would you want a separate PR for 4.6.2 support seems like it should be quite trivial based on this fix ?

PR #362, opened 2016-02-03
Add contracts for System.Drawing.Imaging.EncoderParameters

### LINKED ISSUE #352: Add contracts for System.Drawing.Imaging.EncoderParameters

yaakov-h:
Includes stub for System.Drawing.Imaging.EncoerParameter.

Fixes #352.

### DIFF
--- a/Microsoft.Research/Contracts/System.Drawing/System.Drawing.Imaging.EncoderParameters.cs
+++ b/Microsoft.Research/Contracts/System.Drawing/System.Drawing.Imaging.EncoderParameters.cs
@@ -1,0 +1,58 @@
+// CodeContracts
+//
+// Copyright (c) Microsoft Corporation
+//
+// All rights reserved.
+//
+// MIT License
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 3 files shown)


Microsoft.Research/Contracts/System.Drawing/System.Drawing.Imaging.EncoderParameters.cs
```diff
@@ -0,0 +1,58 @@
...
+    // Summary:
+    //     Initializes a new instance of the System.Drawing.Imaging.EncoderParameters class
+    //     that can contain the specified number of System.Drawing.Imaging.EncoderParameter
+    //     objects.
+    //
+    // Parameters:
+    //   count:
+    //     An integer that specifies the number of System.Drawing.Imaging.EncoderParameter
+    //     objects that the System.Drawing.Imaging.EncoderParameters object can contain.
+    public EncoderParameters(int count) {
```
SergeyTeplyakov: There is no explicit precondition for this class, but there is an implicit one from the array "constructor" - `count >= 0`. Not sure whether this would be helpful or not though.

yaakov-h: I noticed that, I'm not sure what to do about implicit preconditions. For example, `EncoderParameter` (singular) is quite unsafe and can easily be coaxed into throwing a `NullReferenceException`.

SergeyTeplyakov: This is up to you. Suggested precondition could be helpful in some cases for static analysis, but not sure whether they will have practical differences or not.

yaakov-h: I'll leave it then, as the documentation doesn't say anything about constraining this parameter.

Issue #366, opened 2016-02-03
CCDoc still failing


tom-englert:
After installing 1.10.10126.2-rc1 the stack overflow is gone, but now I get this exception:

5>  CCDoc failed with uncaught exception: Object reference not set to an instance of an object.
5>  Stack trace:    at CCDoc.NeutralSourceEmitter.TraverseChildren(IMethodCall methodCall)
5>     at Microsoft.Cci.CodeTraverser.Traverse(IMethodCall methodCall)
5>     at CSharpSourceEmitter.SourceEmitter.TraverseChildren(IBlockExpression blockExpression)
5>     at Microsoft.Cci.CodeTraverser.Traverse(IBlockExpression blockExpression)
5>     at CCDoc.XContract.WriteNeutralFromAST(IExpression expression)
5>     at CCDoc.XPostcondition.WriteTo(XmlWriter writer)
5>     at CCDoc.XmlTraverser.WriteEndElement(XmlReader reader, XmlWriter writer)
5>     at CCDoc.XmlTraverserBase.WriteNodeSingle(XmlReader reader, XmlWriter writer)
5>     at CCDoc.XmlTraverserBase.Transform(XmlReader reader, XmlWriter writer)
5>     at CCDoc.CCDoc.WriteContracts(IDictionary`2 contracts, Options options, DocTracker docTracker)
5>     at CCDoc.CCDoc.RealMain(String[] args)
5>     at CCDoc.CCDoc.Main(String[] args)

tom-englert:
The root cause here is a ```Contract.Ensures(Contract.ForAll(Contract.Result<IList<string>>(), item => item != null));```
With VS2013 generated code CCI decompiles the parameters of ```Contract.ForAll``` as ```ReturnValue``` and ```AnonymousDelegate```, but with VS2015 both parameters are of type ```PopValue```.
There will be probably a better solution than just to ignore these parameters, but at least this fixes the crash.

tom-englert:
Evaluating the generated XML reveals the problem:
VS2013:
```
<ensures 
csharp="System.Diagnostics.Contracts.Contract.ForAll<string>(Contract.Result<System.Collections.Generic.IList<string>>(), (string item) => item != null)" 
vb="System.Diagnostics.Contracts.Contract.ForAll(Contract.Result(Of System.Collections.Generic.IList(Of String))(), (item As String) => item <> Nothing)">for all item in result item != null
</ensures>
```
VS2015 (with above fix):
```
[...]

Issue #277, opened 2015-10-25
Setup test environmet for CCDocGen


tom-englert:
Setup a test environment so we can have tests for CCDocGen, like we have for Clousot.

PR #371, opened 2016-02-05
Add null check to avoid crash when parsing Roslyn generated code.

tom-englert:
To fix https://github.com/Microsoft/CodeContracts/issues/366


SergeyTeplyakov:
Just in case: any chances to add test case for this?

SergeyTeplyakov:
From #366 comment I see that this fix is kind of short-term one. If this is true, I would suggest to put a comment that explains this in the code with (maybe) todo to fix this later properly.

tom-englert:
Yes, this is not a final fix and should not close #366. It just avoids the crash, but now the output is ugly. However without this it's totally unusable due to the crash.
More details to be found in #366.

tom-englert:
About testing: Currently we have no framework for CCDoc tests (#277). Also we do not test Clousot/cccheck with the 4.5 framework/compiler. (#373)

Issue #375, opened 2016-02-09
Can't build CodeContracts - installer building fails


Dennis-Petrov:
Hi everyone.    
I've forked CC from master branch yesterday, and can't build it via batch file. I'm running `buildcc.bat` from MSBuild command prompt (VS2015) and getting this error:

> Build FAILED.
> 
> "D:\Projects\Research\CodeContracts\Microsoft.Research\ManagedContract.Setup\bu
> ildMSI10.xml" (All target) (1) ->
> (CreateVersionFile target) ->
>   EXEC : error : unknown switch '?' [D:\Projects\Research\CodeContracts\Microso
> ft.Research\ManagedContract.Setup\buildMSI10.xml]
>   D:\Projects\Research\CodeContracts\Microsoft.Research\ManagedContract.Setup\b
> uildMSI10.xml(299,5): error MSB3073: The command "git update-index --assume-unc
> hanged ..\common\include\version.cs" exited with code 129.
> 
>     0 Warning(s)
>     2 Error(s)

My machine has git 2.7.0.2 x64 installed.

SergeyTeplyakov:
You need to use VS2013. VS2015 is still not supported (the error message looks weird, but even if this one will pass you'll face another issue with VS2015)..

hubuk:
I was able to reproduce the issue. The error message is indeed weird:
```
  git update-index --assume-unchanged ..\common\include\version.cs
EXEC : error : unknown switch '?' [D:\Github\Microsoft\CodeContracts\Microsoft.Research
\ManagedContract.Setup\buildMSI10.xml]
  usage: git update-index [<options>] [--] [<file>...]

      (...)
      --assume-unchanged    mark files as "not changing"
      (...)
```
When running on dev cmd for VS2013 with the same version of git (2.7.0.2 x64) everything works fine.

Dennis-Petrov:
@SergeyTeplyakov: well... I've built CC about 3 weeks ago successfully from master branch using VS 2015. Moreover, I was able to build it from unpacked ZIP-archive, while now I had to clone git repository locally.

Definitely something were changed in build scripts. Also note, that CC libraries are built just fine. The case is in installer. Are you sure, that the problem is in VS (MSBuild version)?

yaakov-h:
VS2013 Developer Command Prompt plus Git in my path (I think latest Git for Windows) works fine for me, but I am also building from a cloned repository.

[...]

PR #376, opened 2016-02-09
Fixed invalid dash character in build script.

hubuk:
Fix for #375.


yaakov-h:
Do you have to actually run a git command, or can you just use `where git`?

hubuk:
One ```git rev-parse``` command answers two questions:
1. Is git available?
2. Is current directory under git repository?

yaakov-h:
Ah, ok.

PR #387, opened 2016-02-22
Add contracts to System.Windows.Forms, System.Windows.Data classes

Patashu:
Just a bunch of contracts that I noticed were missing.

First pull request; LMK if I did anything wrong.


Patashu:
My pull request was denied due to many lines like

  EXEC : error : Method System.Windows.Forms.DataGridViewRowCollection.AddInternal(System.Boolean, System.Object[]) exists only in contract reference assembly [C:\projects\codecontracts\Microsoft.Research\Contracts\System.Windows
.Forms\System.Windows.Forms10.csproj]

But in the system library, this method does exist:

internal int AddInternal(bool newRow, object[] values)
{
...
}

Can you explain what I am doing wrong/how to fix the problem?

hubuk:
@Patashu The problem is that you are specifying contracts for members which are inaccessible from outside of an assembly, i.e. internals and privates. For example ```internal int AddInternal(bool newRow, object[] values)```. This method is an internal method which can be accessed only by code contained in System.Windows.Forms.dll assembly. Defining contract for such method is of no use for consumers as they have no knowledge (other than reflection) of this method.

So, why the error mentions that the method exist only in contract reference assembly? It clearly exists also in System.Windows.Forms.dll but is not accessible.
That's correct, but the reason is that the check which reported the error is performed against a contract reference assembly and a reference assemblies contained in ```C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework```. These are assemblies which contains only externally visible members of an original framework assemblies and no code. You can check it in ILDasm.

So finally, to make your change work you need to remove all internal and private members/types from your code.

Patashu:
Good explanation, thanks. I'll fix it tomorrow.

PR #393, opened 2016-03-01
Add preconditions to System.Text.Encoding

jeroen-mostert:
The intent is to comply with the public documentation for `System.Text.Encoding` as well as what existing implementations actually do.

As this class is abstract and CC doesn't allow adding preconditions in overrides, this helps those who implement encodings (especially for cases where the precondition is entirely missing but the postcondition is present, like `GetMaxByteCount` and `GetMaxCharCount`).

### DIFF
--- a/Microsoft.Research/Contracts/MsCorlib/System.Text.Encoding.cs
+++ b/Microsoft.Research/Contracts/MsCorlib/System.Text.Encoding.cs
@@ -158,6 +158,8 @@ namespace System.Text
     public virtual string GetString(byte[] bytes, int index, int count)
     {
       Contract.Requires(bytes != null);
+      Contract.Requires(index >= 0 && index <= bytes.Length);
+      Contract.Requires(count >= 0 && index + count <= bytes.Length);
       Contract.Ensures(Contract.Result<string>() != null);
 
       return default(string);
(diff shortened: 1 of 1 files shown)


SergeyTeplyakov:
@yaakov-h Could you please review this stuff. 

Everything looks good for me (checked new contracts with references sources and they looks good).

yaakov-h:
I've had odd issues when combining contracts with `&&` before. It would be better to split them into separate preconditions.

aarondandy:
Those contracts look valid to me.

jeroen-mostert:
@yaakov-h Unless you can qualify the "odd issues", I'm reluctant to break up expressions whose parts semantically belong together, because it hurts readability (`index >= 0 && index <= chars.Length` is instantly recognizable as a range check on `index`; the same expression split in two isn't and invites you to think about the clauses separately when this isn't relevant).

Of course, if it has to be done it has to be done, but I don't want to cargo cult it, especially if it's something that needs to be fixed elsewhere.

yaakov-h:
It's one of the many static-checker-against-VS2015-compiled-code bugs. It can sometimes prove each side of the precondition individually, but cannot prove it when combined, and in some cases even assuming the combination does not satisfy the checker.

PR #400, opened 2016-03-24
Replaced "did you meant" with "did you mean".

davidhesselbom:
The current version reads "Did you meant" in a bunch of places, which isn't proper English and makes Code Contracts look like a work in progress :)

The phrase "Did you mean indexing with x instead of y" still sounds odd, though. Maybe "Did you mean to index with" or "Did you mean index" would work better?


davidhesselbom:
There are several more occurrences of "Did you meant" in `.txt` files in the `Documentation\Papers\OOPSLA2012` folder...

SergeyTeplyakov:
@davidhesselbom For second comment: I'm not sure exactly, but maybe it was just a copy-paste.

Thanks for the fix.

PR #410, opened 2016-05-02
For 64-bit assemblies, preserve the imagebase value when rewriting.

mike-barnett:
Up until now, the .imagebase value written into assemblies was 0x400000. This works for 32-bit assemblies, but not for 64-bit assemblies. So I tried to preserve the value read in on an assembly and use that when writing it. This seems to work for 64-bit assemblies, but for some reason can create a bad 32-bit assembly if the input assembly had a value greater than the default. I tried to debug this, but the header values written by CCI1 and CCI2 are different and it isn't clear to me how they are computed. So I left the old behavior for 32-bit assemblies and have the new behavior only for 64-bit assemblies.

### DIFF
--- a/System.Compiler/Metadata.cs
+++ b/System.Compiler/Metadata.cs
@@ -318,7 +318,7 @@ namespace System.Compiler.Metadata{
     internal int   addressOfEntryPoint;
     internal int   baseOfCode;
     internal int   baseOfData;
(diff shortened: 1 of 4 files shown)


SergeyTeplyakov:
Looks good to me.

jmckone3:
Thanks for taking care of this.

PR #413, opened 2016-05-04
Comment modification for language code

DanAtkinson:
Noticed that the comment for the `lang` on `AddSort` had 'us-en', not en-us.


SergeyTeplyakov:
Thanks a lot for a fix.

PR #445, opened 2016-06-28
Additional fix for #340

tom-englert:
Type arguments can't start with a reserved char, so fall back to standard behavior in this case.


SergeyTeplyakov:
Thanks a lot for fixing it!

PR #446, opened 2016-06-28
Add a test case and revert changes in reader as it breaks the tests.

tom-englert:
After adding a test case I could not reproduce the issue described in #340, but only found that the reader failed all cases with the changes.


SergeyTeplyakov:
Looks good. Thanks.

hickford:
Please could you make a release with this fix? And publish to https://www.nuget.org/packages/DotNet.Contracts/

Issue #450, opened 2016-07-19
Code Contracts Rewriter creates invalid assembly references


Alois-xx:
When an assembly is rewritten the assembly reference flags to all reference assemblies still contain 0x70 in their flags which can cause spurious runtime exceptions in the form of InvalidCastExceptions. 

You need to decompile the rewritten binary with Mono Cecil or similar tools to see the wrongly encoded assembly references for reference assemblies. 

See https://aloiskraus.wordpress.com/2016/07/18/the-non-contracting-code-contracts/ for more information.

Yours,
   Alois Kraus

Daniel-Svensson:
This might explain why we have been getting similar error messages in the Visual Studio designer (causing it to crash).

piedar:
This can be closed now, right?  Would be nice to bump the NuGet package to include the fix.

Alois-xx:
The change should fix the issue.

PR #452, opened 2016-07-26
Forced AssemblyRef flags to be ECMA compatible when writing output

mnkraus:
This pull request is related to issue #450 (https://github.com/Microsoft/CodeContracts/issues/450).
My suggested fix is to force the assembly reference flags to be ECMA compatible just before writing the new assembly


SergeyTeplyakov:
LGTM.

PR #459, opened 2016-09-27
Add flag to generate cccheck RSP at build time

yaakov-h:
This PR adds a new flag to the MSBuild properties, `CodeContractsDeferCodeAnalysis`. If this is set, Code Contracts will do everything it need to do for static analysis except actually run `cccheck.exe`.

This allows developers and automated build processes to manually run the static check at a later time by invoking `cccheck.exe @MyAssemblyName.rsp`.

This option is exposed in the Visual Studio properties pane as 'Create checking artifacts only'. If this option is selected, 'Check in background' and 'Fail build on warnings' are disabled as they are no longer relevant.

This PR also includes minor fixes observed from the BuildCC.bat build log, and renames the RSP files from `MyAssemblyNameccrewrite.rsp`, for example, to `MyAssemblyName.ccrewrite.rsp`.

This has been functionally tested in Visual Studio 2015.

### DIFF
--- a/Foxtrot/Driver/Rewriting/PostRewriteChecker.cs
+++ b/Foxtrot/Driver/Rewriting/PostRewriteChecker.cs
@@ -141,10 +141,7 @@ namespace Microsoft.Contracts.Foxtrot
                 methodName.Matches(ContractNodes.EndContractBlockName))
             {
                 string message = string.Format(Resources.Error_ContractNotRewritten_ContractName_MemberName, method.FullName, this.visitedMembers.Peek().FullName);
-                if (this.handleError != null)
-                {
-                    this.handleError(new Error(1080, message, binding.SourceContext));
-                }
+                this.handleError(new Error(1080, message, binding.SourceContext));
             }
         }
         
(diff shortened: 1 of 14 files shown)


Foxtrot/Driver/Rewriting/PostRewriteChecker.cs
```diff
@@ -141,10 +141,7 @@ public override void VisitMemberBinding(MemberBinding binding)
                 methodName.Matches(ContractNodes.EndContractBlockName))
             {
                 string message = string.Format(Resources.Error_ContractNotRewritten_ContractName_MemberName, method.FullName, this.visitedMembers.Peek().FullName);
-                if (this.handleError != null)
-                {
-                    this.handleError(new Error(1080, message, binding.SourceContext));
-                }
+                this.handleError(new Error(1080, message, binding.SourceContext));
```
SergeyTeplyakov: Not clear from this review: does it really guaranteed that handleError is not null?

yaakov-h: There's an invariant on it as non-null. It's also only set from the constructor, which has a precondition for non-null.

PR #464, opened 2016-10-11
Support SQL Local DB 2016

kkm000:
In my SQLLocalDB-2016-only setup (this is what you get by installing VS2015 Update 3 on a fresh machine), CC fails to connect to default cache. I tracked down to this; I am not sure if you would consider the current approach tenable. I would rather try to connect to `(LocalDb)\MSSQLLocalDB` to see what happens. But I think this is how CC worked pre- b4b89e6a and there was a reason it was changed to be like it is now.

I kinda fixed it by creating an empty "12.0" subkey there, but this is just a hotwired workaround, mostly to confirm that I did find the correct spot.

/cc @tom-englert


SergeyTeplyakov:
Thanks! Looks good.

PR #470, opened 2016-12-01
Add missing contracts found when aligning CC with R#

tom-englert:
When aligning CC with R# External Annotations (https://github.com/JetBrains/ExternalAnnotations), I have identified and added these missing contracts.


SergeyTeplyakov:
Thanks. Looks good.

Issue #454, opened 2016-08-12
Cant install Nuget for DotNet.Contracts 1.10.20606.1


xmarshal:
Could not install package 'DotNet.Contracts 1.10.20606.1'. You are trying to install this package into a project that targets '.NETFramework,Version=v4.6', but the package does not contain any assembly references or content files that are compatible with that framework. For more information, contact the package author.
fo 4.0, 4.6.1, 4.6.2 same problem.

tom-englert:
What version of VS/Nuget are you using?

xmarshal:
VS 2015 ver. 14.0.25425.01 Update 3.
NUget ver. 3.4.4.1321.

Issue #451, opened 2016-07-26
VS15 (Preview) support


yaakov-h:
What needs to be done to add support for VS15, however preliminary.

At minimum we need MSBuild targets for v15.0 (`C:\Program Files (x86)\Microsoft\Contracts\MsBuild`).

What would I need to modify to add such targets to the project?

- [ ] MSBuild targets for MSBuild 15.0
- [ ] Plugin / Project Preferences sheet
- [ ] Tests?
- [ ] Anything else?

yaakov-h:
Note to self: Looks like most of the targets work is in WiX: https://github.com/Microsoft/CodeContracts/blob/79a04c96679b6d400c286dd62d6a2ec77462da72/Microsoft.Research/ManagedContract.Setup/ManagedContracts.wxs#L1078-L1087

tom-englert:
I had to setup a build server with VS15 for evaluation. 
The only thing needed to get it run was to copy
`C:\Program Files (x86)\MSBuild\14.0\Microsoft.Common.Targets\ImportAfter\CodeContractsAfter.targets`
to 
`C:\Program Files (x86)\MSBuild\15.0\Microsoft.Common.Targets\ImportAfter\CodeContractsAfter.targets`

Of course you don't get the CC settings page in the project properties, but it's sufficient to try out existing solutions

yaakov-h:
That works for me too, though I haven't tried CC with any of the new C#7 features yet.

I think the settings pane should show up if you run the MSI after installing VS15. If not, there's a breaking change somewhere in VS.

sickboy:
Location since recent builds seem to be 
> C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Microsoft.Common.targets\ImportAfter

takinosaji:
@patroza doesnt work in release version. Is the content of targets file supposed to be changed?

mmusaev:
Does this mean that Code Contracts will be supported in Visual Studio 2017 soon?

patroza:
@takinosaji I've decided to move on and not have this headache at every new VS version :)

yaakov-h:
@mmusaev Depends what you mean by "supported", and if anyone's left alive here to merge #482.

mmusaev:
I meant by support if Code Contract would be integrated into Visual Studios 2017 Settings Tab.

yaakov-h:
I honestly have no idea how to do that at the moment.

[...]

Issue #368, opened 2016-02-04
MSI installations conflicts with NuGet


yaakov-h:
As mentioned in #13, the NuGet package does not automatically install the targets file in the project.

If I add the following import after `Microsoft.CSharp.targets`:
```
  <PropertyGroup>
    <CodeContractsInstallDir>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)..\packages\DotNet.Contracts.1.10.10126.4\'))</CodeContractsInstallDir>
  </PropertyGroup>
  <Import Project="$(CodeContractsInstallDir)\MsBuild\v$(VisualStudioVersion)\Microsoft.CodeContracts.targets"/>
```

I get the following build error:

Severity | Code | Description | Project | File | Line | Suppression | State
--- | --- | --- | --- | --- | --- | --- | ---
| Error | "obj\Debug\Decl\ContractNuGetTest.dll;obj\Debug\Decl\ContractNuGetTest.dll" is an invalid value for the "OutputAssembly" parameter of the "Csc" task. Multiple items cannot be passed into a parameter of type "Microsoft.Build.Framework.ITaskItem". | ContractNuGetTest |

Judging by MSBuild diagnostic output, `Microsoft.CodeContracts.targets` from the MSI is being included, which means that the targets file from NuGet comes in afterwards and is adding items to now-not-empty ItemGroups.

If I edit the `csproj` file to reflect the following:
```
  <PropertyGroup>
    <DontImportCodeContracts>True</DontImportCodeContracts>
  </PropertyGroup>
  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
  <PropertyGroup>
    <CodeContractsInstallDir>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)..\packages\DotNet.Contracts.1.10.10126.4\'))</CodeContractsInstallDir>
  </PropertyGroup>
  <Import Project="$(CodeContractsInstallDir)\MsBuild\v$(VisualStudioVersion)\Microsoft.CodeContracts.targets"/>
```

Then Contracts work from NuGet, overriding any MSI-installed Code Contracts on the system.

So, I see three problems:

1. Contracts from NuGet aren't automatically set up
2. Contracts from the MSI take precedence over Contracts from NuGet. I'm not sure if this one is solvable.
3. Contracts from NuGet conflict with Contracts from the MSI, it isn't a clean replacement.

I think the simplest solution to this is:

[...]

Issue #313, opened 2015-11-26
NuGet package refuses to install


NameOfTheDragon:
The NuGet package refuses to install (under VS2015) giving the reason:
 > Could not install package 'Microsoft.Contracts 1.9.1.1004'. You are trying to install this package into a project that targets '.NETFramework,Version=v4.6', but the package does not contain any assembly references or content files that are compatible with that framework. For more information, contact the package author.

Note: that is a private build, I'm building from the master branch on my TeamCity server using the supplied scripts. The built NuGet package is available from my build server at http://build.teamserver.tigranetworks.co.uk/project.html?projectId=OpenSource_MicrosoftCodeContracts 

I have found that NuGet requires at least one item to match the target framework in either the content or lib directory, otherwise it'll refuse to install the package. VS2015 is using NuGet 3.0 but I have had similar issues on my TeamCity server which is using NuGet 2.8. Either way, the package in its current form refuses to install under VS2015.

NameOfTheDragon:
By the way, your versioning scheme `1.9.1MMdd` is terrible as it results in version numbers going backwards on 1st January.

NameOfTheDragon:
This might be a NuGet issue: https://github.com/NuGet/Home/issues/1710

PR #473, opened 2016-12-30
Fix nuget package

Daniel-Svensson:
This PR makes it possible to use the code contracts nuget package by just adding Dotnet.Contracts as a nuget reference. (first commit)
It also makes some minor improvements to the target files such as adding support for .Net 4.6.2 (second commit). This can be moved to a second PR if requested.

## Nuget improvements
The .targets files will be automatically included in the Project and no further action is required unless the MSI is also installed .

It should solve the following issues
* #313 Nuget package refuse to install 
* #454 Cant install Nuget for DotNet.Contracts 1.10.20606.1
* #13 CodeContracts as NuGet package

It should also allowing code contract rewriter to run for VS 2017 #451 

However *if you have installed the MSI* you do need to *include the following workaround at the top of your project file* (it must be before "Microsoft.CSharp.targets" is included) which is mentioned in Issue #368 

 ```
  <PropertyGroup>
    <DontImportCodeContracts>True</DontImportCodeContracts>
  </PropertyGroup>
```

## .target file improvements 

* Remove extra space from library path
 Use relative path when including analysis targets so that the same file can be used for multiple MsBuild versions

### DIFF
--- a/Microsoft.Research/ManagedContract.Setup/DotNet.Contracts.nuspec
+++ b/Microsoft.Research/ManagedContract.Setup/DotNet.Contracts.nuspec
@@ -12,6 +12,7 @@
     <licenseUrl>https://github.com/Microsoft/CodeContracts/blob/master/LICENSE.txt</licenseUrl>
   </metadata>
   <files>
-     <file src="**\*" />
+     <file src="**\*" target="build\" />
+	 <file src="..\..\..\..\..\..\Dotnet.Contracts.targets" target="build\" />
   </files>
 </package>
--- a/Microsoft.Research/ManagedContract.Setup/DotNet.Contracts.nuspec
+++ b/Microsoft.Research/ManagedContract.Setup/DotNet.Contracts.nuspec
@@ -13,6 +13,6 @@
   </metadata>
   <files>
      <file src="**\*" target="build\" />
-	 <file src="..\..\..\..\..\..\Dotnet.Contracts.targets" target="build\" />
+     <file src="..\..\..\..\..\..\Dotnet.Contracts.targets" target="build\" />
   </files>
 </package>
--- a/Microsoft.Research/ManagedContract.Setup/DotNet.Contracts.targets
+++ b/Microsoft.Research/ManagedContract.Setup/DotNet.Contracts.targets
@@ -1,0 +1,14 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <CodeContractsInstallDir>$(MSBuildThisFileDirectory)</CodeContractsInstallDir>
+  </PropertyGroup>
+
+  <!-- Use MsBuild specific targets file if availible -->
+  <Import Project="$(CodeContractsInstallDir)\MsBuild\v$(MSBuildToolsVersion)\Microsoft.CodeContracts.targets"
+   Condition="Exists('$(CodeContractsInstallDir)\MsBuild\v$(VisualStudioVersion)\Microsoft.CodeContracts.targets')" />
+  <!-- Fallback to Version 14 in hope that it is the most up to date version otherwise --> 
+   <Import Project="$(CodeContractsInstallDir)\MsBuild\v14\Microsoft.CodeContracts.targets"
+   Condition="!Exists('$(CodeContractsInstallDir)\MsBuild\v$(MSBuildToolsVersion)\Microsoft.CodeContracts.targets')" />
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: Microsoft.Research/ManagedContract.Setup/DotNet.Contracts.nuspec [hunk 1]
```diff
@@ -12,6 +12,7 @@
     <licenseUrl>https://github.com/Microsoft/CodeContracts/blob/master/LICENSE.txt</licenseUrl>
   </metadata>
   <files>
-     <file src="**\*" />
+     <file src="**\*" target="build\" />
+	 <file src="..\..\..\..\..\..\Dotnet.Contracts.targets" target="build\" />
```
yaakov-h: Indentation seems off on this line.

Microsoft.Research/ManagedContract.Setup/DotNet.Contracts.targets
```diff
@@ -0,0 +1,14 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <CodeContractsInstallDir>$(MSBuildThisFileDirectory)</CodeContractsInstallDir>
+  </PropertyGroup>
+
+  <!-- Use MsBuild specific targets file if availible -->
+  <Import Project="$(CodeContractsInstallDir)\MsBuild\v$(MSBuildToolsVersion)\Microsoft.CodeContracts.targets"
```
yaakov-h: Why mix `MSBuildToolsVersion` with `VisualStudioVersion`?

Microsoft.Research/ManagedContract.Setup/DotNet.Contracts.nuspec
Daniel-Svensson: probably tabs instead of spaces.
Will fix

Microsoft.Research/ManagedContract.Setup/DotNet.Contracts.targets
Daniel-Svensson: Seems like I missed to change it. It was originally VisualStudioVersion from another .targets file but MSBuildTools seemed correct. Will fix

Microsoft.Research/ManagedContract.Setup/DotNet.Contracts.targets
```diff
@@ -0,0 +1,14 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <CodeContractsInstallDir>$(MSBuildThisFileDirectory)</CodeContractsInstallDir>
```
Igorbek: does it set it unconditionally? A condition could be added in order to be able to override the value in the project file.

yaakov-h: Why would you want to override the value for "where is Code Contracts" to something other than "right here" when using the NuGet package?

Daniel-Svensson: I am of the same opinion as yaakov-h, I think that you add the nuget package in order to use that specific version and no other. 

We could relatively simple make it use the system installed version if one is installed and ignore the nuget version, (easier than the other way around), but I would find that un-intuitive and confusing.

yaakov-h: That's backwards from how the package would typically be used.

If you don't have CC installed, the package should use it's own version.

If you do have CC installed and no package, you would use the system version.

If you have CC installed and you have projects that use different package versions, each project should use the version that it requires, or was last tested with, or whatever. Modifying the system version should introduce changes to a project that has taken a dependency on a _specific version_ of the NuGet package.

I see absolutely no point in having system Code Contracts installed, then installing a NuGet package, and having the package just defer to the global installation that was already there and being used. The package becomes redundant and confusing.

Igorbek: Ok, agreed :)

jeremysimmons:
@Daniel-Svensson I would really love to see this completed. Can you sign the contributor license agreement?

Daniel-Svensson:
@jeremysimmons i could try looking in a week or two if required (I am a bit unsure this is complex enough to warrant signing a cla).  I did not find any info about any CLA in the main readme.md do you know where to do so for this project?

Daniel-Svensson:
Closing since the project seems abandoned

PR #475, opened 2016-12-30
Target correct reference assemblies for .Net 4.6.2 and 4.5.2

Daniel-Svensson:
* Add support for .Net 4.6.2 
** Any possible future 4.6.x release would also choose correct reference assemblies
** Fixes #475 
* Add support for .Net 4.5.2 for toolset version 12

Default TargetFramework (if target framework is missing) changed from 3.5 to 4.0


Daniel-Svensson:
Closing since project seems anandoned

PR #482, opened 2017-03-08
Add MSBuild targets for VS2017

yaakov-h:
This PR adds MSBuild targets for MSBuild 15.0 (Visual Studio 2017).

It also uses the .NET 4.6 Contract BCLs when targeting .NET 4.6.2.

I haven't done anything on the IDE integration side, partly because Visual Studio 2017 has moved a whole lot of cheeses and it's not a simple case of incrementing "14" to "15" and calling it a day.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ Microsoft.Research/CCDoc/Sandcastle/Sandcastle.zip
 *.user
 *.userosscache
 *.sln.docstates
+.vs
 
 # Build results
 [Dd]ebug/
--- a/Microsoft.Research/ManagedContract.Setup/MSBuild/v12.0/Microsoft.CodeContracts.targets
+++ b/Microsoft.Research/ManagedContract.Setup/MSBuild/v12.0/Microsoft.CodeContracts.targets
@@ -70,6 +70,11 @@
             <CodeContractsReferenceAssemblyLibPath>$(CodeContractsInstallDir)Contracts\.NETFramework\v4.6</CodeContractsReferenceAssemblyLibPath>
           </PropertyGroup>
         </When>
+        <When Condition="'$(TargetFrameworkVersion)' == 'v4.6.2'">
+          <PropertyGroup>
+            <CodeContractsReferenceAssemblyLibPath>$(CodeContractsInstallDir)Contracts\.NETFramework\v4.6</CodeContractsReferenceAssemblyLibPath>
+          </PropertyGroup>
+        </When>
         <Otherwise>
           <PropertyGroup>
             <CodeContractsReferenceAssemblyLibPath>$(CodeContractsInstallDir)Contracts\v3.5</CodeContractsReferenceAssemblyLibPath>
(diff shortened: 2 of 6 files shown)


ceddlyburge:
Hi there

Is this PR released? 

I would like my Contracts to start throwing exceptions so that my tests can pass in VS 2017, which I think this does.

I'm not too worried about static analysis and suchlike.

It might be worth telling Microsoft that this is stopping me buying VS 2017 at the moment, to see if they can offer any help.

Cheers

Cedd

yaakov-h:
I've been using https://github.com/WiseTechGlobal/CodeContracts/tree/wtg/rel/v1.10.20308.4 (which includes this PR) for over a month and it's quite stable. A few things aren't working in VS2017 including the properties pane, code snippets and background static analysis, but the rest is working just fine, including the rewriter.

You can compile it yourself, or when I next get into the office I can upload a nupkg from that fork.

ceddlyburge:
Thanks Yaakov.

artkpv:
Hi, thanks for tips!

Is there way to install package into VS2017 project so that CC will be used at minimum (replacing Contract.Requires/Assert/..) ? 

I've tried https://github.com/WiseTechGlobal/CodeContracts . But it fails with:

```
PM> Install-Package [PATH]\src\external\DotNet.Contracts.1.10.20308.4.nupkg


Attempting to gather dependency information for package 'DotNet.Contracts.1.10.20308.4' with respect to project 'Gameo.GE.Web', targeting '.NETFramework,Version=v4.5.2'
Install-Package : An error occurred while retrieving package metadata for 'DotNet.Contracts.1.10.20308.4' from source '...y\src\external'.
  '...\src\external\DotNet.Contracts.1.10.20308.4.nupkg' is not a valid nupkg file. Error: 'End of Central Directory record could not be found.'.
  End of Central Directory record could not be found.
At line:1 char:1
+ Install-Package ...src\external ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Install-Package], Exception
    + FullyQualifiedErrorId : NuGetCmdletUnhandledException,NuGet.PackageManagement.PowerShellCmdlets.InstallPackageCommand
 
Time Elapsed: 00:00:00.2067542
PM> 
```

yaakov-h:
Sounds like you have a corrupted zip file.

artkpv:
@yaakov-h 
Yes, it was broken. But now it needs different .NET version. Arrrgh! )

```
Could not install package 'DotNet.Contracts 1.10.20308.4'. You are trying to install this package into a project that targets '.NETFramework,Version=v4.5.2', but the package does not contain any assembly references or content files that are 
compatible with that framework. For more information, contact the package author.
```

Update. [This package](https://www.nuget.org/packages/CodeContracts.MSBuild/) seems to work (as per generated IL code) but I'm not sure about it (can we trust it?).

PR #496, opened 2017-09-25
Bugfix/495 dotnet build

### LINKED ISSUE #495: Build with CodeContracts fails with dotnet build
By manually adding imports and msbuild properties (b/c no IDE support in VS 2017), ccrewrite works on netcoreapp and netstandard projects (only when [portable PDBs](https://github.com/dotnet/core/blob/master/Documentation/diagnostics/portable_pdb.md) are not used).

Running the build using `msbuild /t:build` works. Running the build using `dotnet build` fails:

```
dotnet build /v:normal
Build started 2017-09-25 9:06:34 AM
...
       CreateCodeContractReferenceAssembly:
         "C:\Program Files (x86)\Microsoft\Contracts\Bin\ccrefgen.exe" "@obj\Debug\netstandard2.0\MyProject.ccrefgen.rsp"
         elapsed time: 881.9998ms
       CodeContractRewrite:
         "C:\Program Files (x86)\Microsoft\Contracts\Bin\ccrewrite.exe" "@MyProject.ccrewrite.rsp"
                There were errors reported in Tv.Web.Upfront.Model's metadata.
                Exception of type 'System.OutOfMemoryException' was thrown.
     1>ccrewrite : error : Foxtrot:     There were errors reported in Tv.Web.Upfront.Model's metadata. [C:\src\my\project\MyProject.csproj]
                Exception of type 'System.OutOfMemoryException' was thrown.
         elapsed time: 61.4774ms
     1>C:\Program Files (x86)\Microsoft\Contracts\MsBuild\v15.0\Microsoft.CodeContracts.targets(275,5): error MSB3073: The command ""C:\Program Files (x86)\Microsoft\Contracts\Bin\ccrewrite.exe" "@MyProject.ccrewrite.rsp"" exited with code -1. [C:\src\my\project\MyProject.csproj]
     1>Done Building Project "C:\src\my\project\MyProject.csproj" (Build target(s)) -- FAILED.

Build FAILED.
```

The root cause is this block of the `Microsoft.CodeContracts.targets` file:
https://github.com/Microsoft/CodeContracts/blob/master/Microsoft.Research/ManagedContract.Setup/MSBuild/v15.0/Microsoft.CodeContracts.targets#L292

Target `GetFrameworkSdkPath`, and `sn.exe` don't exist in the dotnet build system.

johncrim:
This commit fixes #495 , which I opened. I believe it's reasonable to skip the sn.exe step when building under dotnet, [since strong-naming is discouraged for non-.NET Core SDK libraries and apps](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/strong-name-signing.md).


yaakov-h:
If sn.exe doesn’t exist in in .net core sdk, how does it do strong-name signing (on Windows)?

johncrim:
From the examples I've seen (eg https://github.com/aspnet/Logging/blob/dev/Directory.Build.props ), sn.exe is run outside of the `msbuild blah.csproj` or `dotnet build` process.

yaakov-h:
Lines 8 and 9 in that file suggest otherwise.

johncrim:
@yaakov-h : My last comment didn't provide a useful answer. `Directory.Build.props` defines an msbuild property reference to the .snk file, and properties in that file become part of all the msbuild file in project subdirectories.

However, none of the above answers "how is strong-naming done in dotnet core projects". I haven't yet been able to find an answer to that, but indications are that strong-naming isn't done by the default dotnet msbuild targets. This topic deserves further research before merging my fix.

[Framework NuGet package servicing](https://blogs.msdn.microsoft.com/dotnet/2014/01/22/net-4-5-1-supports-microsoft-security-updates-for-net-nuget-libraries/) and the GAC don't apply to .NET Core on non-Windows OSes.

